Issue #4, opened 2018-03-08
makewin**chk.cmd implicit depended VS2017 Enterprise


josephsmeng:
in both makewin32chk.cmd and makewin64chk.cmd line 6,
the  call function are VS2017 Enterprise only, and makes  the
community version to fail.

I think it would be better to do some file check here to dectct
which version is actually installed on current machine.

msftphil:
Do you have a pull request in mind that I can take a look at?

tabs-not-spaces:
perhaps a solution would be to improve the programatic nature of the make script by using powershell? if so, I'm happy to put in  PR with an example that works for me...

msftphil:
Yea, I think that's a totally reasonable way to go.  Looking forward to seeing a PR!

AndrewPardoe:
Visual Studio 2017 includes a tool called VSWhere that is used to locate installations of Visual Studio. I would suggest using that tool. For example, here's my VCVars.cmd: it finds the latest installation of VS and calls the associated VCVars: https://gist.github.com/AndrewPardoe/689a3b969670787d5dba538bb0a48a1e.

tabs-not-spaces:
Good idea Andrew, however if it doesn't exist for earlier versions, then it's best not to rely on it. I'm putting in a PR shortly.

tabs-not-spaces:
have a look at the latest PR - needs testing by someone with VS Community, but should work seamlessly.
[https://github.com/Microsoft/msix-packaging/pull/6](url)

AndrewPardoe:
VSWhere works for all versions of VS2017, and has been installed with the VS installer since VS2017 Update 1.

PR #6, opened 2018-03-09
build script for windows now in powershell and works with VS 2017 community & enterprise

tabs-not-spaces:
as discussed in [https://github.com/Microsoft/msix-packaging/issues/4](url)

Script uses a simple function to look at a few logical places where the vs dev cmd scripts might be living. No external pre-reqs. Has not been tested with community, so please clone and confirm.

Script defaults to 64bit build - use the "-architecture" flag to define the arch if you want to change it.

### DIFF
--- a/makewin.ps1
+++ b/makewin.ps1
@@ -1,0 +1,70 @@
+[CmdletBinding()]
+param(
+    [parameter()]
+    [ValidateSet(, "64-bit", "32-bit")]
+    [string]$Architecture = "64-bit"
+)
+function Set-VsDevCmd {
+    <#
+    .SYNOPSIS
+        Finds the VSDevCmd // VCVars64 scripts and adds the params to the environment in powershell
+    .PARAMETER Architecture
+        Set your system architecture to either 32-Bit or 64-Bit
+    .NOTES
+        Kudos https://stackoverflow.com/questions/2124753/how-can-i-use-powershell-with-the-visual-studio-command-prompt
+    #>
+    [CmdletBinding()]
+    Param(
+        [parameter()]
+        [ValidateSet(, "64-bit", "32-bit")]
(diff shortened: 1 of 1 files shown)


msftphil:
@AndrewPardoe, is there another mechanism that the script can use to determine the correct (as opposed to the default) install path and take the latest version of the tooling?

@tabs-not-spaces, this script looks like it builds in-place in-tree as opposed to a subdirectory.  Is that intentional?

tabs-not-spaces:
@AndrewPardoe - does the VSWhere tool list all install paths in a system with multiple versions of VS installed? if so, I'm happy to modify to use that.

@msftphil - not intentional, just how I got it working - i'm assuming I just need to modify the cmake command to change the installation path?

mqmashwani:
@tabs-not-spaces Please look at the Microsoft project [vssetup.powershell](https://github.com/microsoft/vssetup.powershell).

This Powershell module contains the command "Get-VSSetupInstance". This command will detect all instances of Visual Studio 2017. Using this module, side by side installations and custom install paths should not be a concern. The results can be filtered as appropriate for your use case.

image


For earlier versions the best detection methods are referenced at [MZTools](https://www.mztools.com/articles/2008/MZ2008003.aspx). Depending on which registry keys are searched additional information about the installation, such as the install path, can be found from searching the properties of the listed keys.

nschonni:
Here is an example of using the VSWhere to find a default install https://github.com/Microsoft/process-customization-scripts/pull/10/files

tabs-not-spaces:
Thanks @nschonni - I've been very time-poor recently, but I'll implement this in shortly and check it in for review

c-don:
We will update the documentation to inform developers to open the developer command prompt.

PR #38, opened 2019-02-13
Fix memory corruption due to TOKEN_PRIVILEGES not allocating enough memory

rudyhuyn:
By default, `TOKEN_PRIVILEGES` can only host 1 `LUID_AND_ATTRIBUTES`, but Win7MsixInstaller writes 2 of them, overwriting the 12 following bytes.

See documentation:

> Important  The constant ANYSIZE_ARRAY is defined as 1 in the public header Winnt.h. To create this array with more than one element, you must allocate sufficient memory for the structure to take into account additional elements.

https://docs.microsoft.com/en-us/windows/desktop/api/winnt/ns-winnt-_token_privileges

### DIFF
--- a/preview/Win7Msix/Win7MSIXInstaller/RegistryDevirtualizer.cpp
+++ b/preview/Win7Msix/Win7MSIXInstaller/RegistryDevirtualizer.cpp
@@ -23,7 +23,7 @@ HRESULT RegistryDevirtualizer::Run(_In_ bool remove)
 {
     std::wstring rootPath = m_loadedHiveKeyName + L"\\Registry";
     RETURN_IF_FAILED(m_rootKey.Open(HKEY_USERS, rootPath.c_str(), KEY_READ));
-    
+
     for (auto mapping : mappings)
     {
         TraceLoggingWrite(g_MsixTraceLoggingProvider,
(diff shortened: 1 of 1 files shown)


preview/Win7Msix/Win7MSIXInstaller/RegistryDevirtualizer.cpp
```diff
@@ -1,412 +1,412 @@
...
+    if (!LookupPrivilegeValue(
+        NULL,            // lookup privilege on local system
+        SE_BACKUP_NAME,   // privilege to lookup 
+        &seBackupLuid))        // receives LUID of privilege
+    {
+        RETURN_IF_FAILED(HRESULT_FROM_WIN32(GetLastError()));
+    }
+
+    PTOKEN_PRIVILEGES pTokenPrivileges = NULL;
+    TOKEN_PRIVILEGES oldTokenPrivileges{};
```
jyvenugo: I noticed that I had to modify this such that, both pTokenPrivileges and oldTokenPrivileges are allocated in the same way initially for app installation to work.

rudyhuyn: `oldTokenPrivileges` is never used, we should remove it.

rudyhuyn:
`oldTokenPrivileges` removed!

PR #39, opened 2019-02-15
Make win7MsixInstaller compatible with self-signed packages.

rudyhuyn:
Add a parameter to install a package using a self-signed certificate. 
Make it possible to retrieve the publisher name when `option == MSIX_VALIDATION_OPTION_ALLOWSIGNATUREORIGINUNKNOWN`.
- Fix segfault when no certificates (CA and self-signed excluded) are found.


rido-min:
What's the expected behavior? and how it compares to Appinstaller App signing requirements? for the next scenarios:

- Unsigned Packages
- Packages signed with certitificate issued by untrusted CAs
- Packages signed with self-signed certificates untrusted
- Packages signed with self-signed certificates trusted (*)

(*)What are the requirements for trusting certs? LocalMachine/TrustedPeople?

rudyhuyn:
Current code:
* Unsigned Packages ==> win7msixInstaller crashes because no certificates found and the code didn't check that `GetCertContext` can return NULL.
* Packages signed with certificate issued by untrusted CAs => Should be tested
* Packages signed with self-signed certificates untrusted => win7msixinstaller will ignore the self-signed certificate and `GetCertContext` will return NULL, crashing the app
* Packages signed with self-signed certificates trusted (*) => win7msixinstaller will ignore the self-signed certificate and `GetCertContext` will return NULL, crashing the app

This diff (with the parameter):
* Unsigned Packages ==> OK
* Packages signed with certificate issued by untrusted CAs => OK
* Packages signed with self-signed certificates untrusted => OK
* Packages signed with self-signed certificates trusted (*) => OK

For `win7msixInstaller` (legacy installation), we should question the necessity of verifying whether the certificate is valid or not, we can display a warning but not especially block the installation

rido-min:
Thanks for the update, can you elaborate on how the "parameter" will be set by users?

@jvintzel is it ok to install packages unsigned and/or untrusted?

rudyhuyn:
The target is not users but developers, to easily test their .msix. If you create a desktop bridge package for example via VS and install the certificate (double-click or via the powershell script), the .msix can be installed via when you click on it, but not via win7msixinstaller because `SignatureValidator` rejects the self-signed certificate created by visual studio. 

Here is the msix and cert I use for test:
[SampleToast_1.0.9.0_AnyCPU.zip](https://github.com/Microsoft/msix-packaging/files/2887138/SampleToast_1.0.9.0_AnyCPU.zip)

jvintzel:
> The target is not users but developers, to easily test their .msix. If you create a desktop bridge package for example via VS and install the certificate (double-click or via the powershell script), the .msix can be installed via when you click on it, but not via win7msixinstaller because SignatureValidator rejects the self-signed certificate created by visual studio.
> Here is the msix and cert I use for test:
> SampleToast_1.0.9.0_AnyCPU.zip

we have an issue with self-signed certificates, we are working to get that fixed.

rido-min:
>we have an issue with self-signed certificates, we are working to get that fixed.

@jvintzel can you provide more details, or issue number, to track the self-signed certificates issue?

msftrubengu:
Allowing the validation of a self-signed thing to succeed so long as the caller is okay with the unknown signature origins is fine. @jvintzel This change solves the problem we have with self-signed certificates.

wcheng-msft:
If we make self-signed certs that are trusted by default OK to install, would we still need/want the allowsignatureoriginunknown flag?

src/msix/PAL/Signature/Win32/SignatureValidator.cpp
```diff
@@ -152,49 +152,58 @@ namespace MSIX
     static bool GetEnhancedKeyUsage(PCCERT_CONTEXT pCertContext, std::vector<std::string>& values)
     {                   
         //get OIDS from the extension or property
-
+        if (pCertContext == NULL)
+            return false;
```
wcheng-msft: style-wise can we always have braces?

src/msix/PAL/Signature/Win32/SignatureValidator.cpp
```diff
@@ -351,29 +360,40 @@ namespace MSIX
...
         }
-        return pCertContext;
     }
     
     static bool DoesSignatureCertContainStoreEKU(_In_ byte* rawSignatureBuffer, _In_ ULONG dataSize)
     {
-        unique_cert_context certificateContext(GetCertContext(rawSignatureBuffer, dataSize));        
+        unique_cert_context certificateContext(GetCertContext(rawSignatureBuffer, dataSize, false));
+        if (certificateContext.get() == NULL)
+            return false;
```
wcheng-msft: add braces to if statement
```suggestion
        {
            return false; 
        }
```

src/msix/PAL/Signature/Win32/SignatureValidator.cpp
```diff
@@ -323,7 +332,7 @@ namespace MSIX
         return(true);
     }
 
-    static PCCERT_CONTEXT GetCertContext(BYTE *signatureBuffer, ULONG cbSignatureBuffer)
+    static PCCERT_CONTEXT GetCertContext(BYTE *signatureBuffer, ULONG cbSignatureBuffer, bool bAllowUnknownSignature)
```
wcheng-msft: I think the intent to skip selfsigned certificates is for the StoreEKU check-- store signed packages should not be self signed. However GetPublisherName should always be OK to handle self-signed certs regardless of options.

Also, don't use Hungarian on the variable name.

So, I would suggest instead explicitly naming it bool allowSelfSignedCert, and have the StoreEKU caller pass in false, and the GetPublisherName pass in true.
If changed that way, we don't need/want to allow the option AllowUnknownSignature because self-signed certs would simply be allowed if they chained to a trusted root.

src/msix/PAL/Signature/Win32/SignatureValidator.cpp
```diff
@@ -351,29 +362,42 @@ namespace MSIX
...
         {   //pkcs7 -- get the end entity
+            PCCERT_CONTEXT pCertContext = NULL;
             while (NULL != (pCertContext = CertEnumCertificatesInStore(certStoreHandle.get(), pCertContext)))
-            {   if (IsCertificateSelfSigned(pCertContext, pCertContext->dwCertEncodingType, 0) || IsCACert(pCertContext))
-                { continue;
+            {    
+                if (IsCertificateSelfSigned(pCertContext, pCertContext->dwCertEncodingType, 0))
+                {
+                    if (allowSelfSignedCert)
+                        return pCertContext;
```
wcheng-msft:
```suggestion
                    {
                        return pCertContext;
                    }
```

rudyhuyn:
I was not expecting a review of the new diff that fast! 

I agree with you, if there is a plan to fix self-signed certificate we should not add this extra parameter, I removed it but I kept the other changes in `SignatureValidator`.

Thanks to everyone for your feedback!

PR #49, opened 2019-03-02
[Win7MsixInstaller] Fix UI drawing when the dialog is minimized or partially out of the screen.

rudyhuyn:
The title, description as well as the application logo were never redraw when the window was partially out of the screen or after being minimized.

The fix splits `DisplayPackageInfo` into 2 parts:
- `LoadInfo`: called only one time, extracting application name, publisher name, version number and logo and caching them to improve performance.
- `DrawPackageInfo, called when the window receives the event `WM_PAINT`, redrawing textblocks and the logo

+ Load the progress bar only when we will display it.

**Before:**

**Fixed:**

### DIFF
--- a/preview/Win7Msix/Win7MSIXInstaller/InstallUI.cpp
+++ b/preview/Win7Msix/Win7MSIXInstaller/InstallUI.cpp
@@ -60,14 +60,25 @@ HRESULT GetStreamFromFile(IAppxPackageReader* package, LPCWCHAR name, IStream** 
 //
 // PURPOSE: This compiles the information displayed on the UI when the user selects an msix
 //
-// windowText: pointer to a wstring that the window message will be saved to
-HRESULT UI::DisplayPackageInfo(HWND hWnd, RECT windowRect)
-{
-    auto displayText = L"Install " + m_displayName + L"?";
-    auto messageText = L"Publisher: " + m_publisherCommonName + L"\nVersion: " + m_version;
-	CreateProgressBar(hWnd, windowRect, m_numberOfFiles);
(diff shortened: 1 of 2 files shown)


preview/Win7Msix/Win7MSIXInstaller/InstallUI.cpp
```diff
@@ -464,17 +475,15 @@ BOOL ChangeText(HWND parentHWnd, std::wstring displayName, std::wstring messageT
...
-    }
-        
-    if (g_image != nullptr)
-    {
-        Gdiplus::Status status = graphics.DrawImage(g_image, g_width - 200, 25);
-    }
+	if (logoStream != nullptr)
+	{
+		// We shouldn't fail if the image can't be loaded, just don't show it.
+		auto g_image = Gdiplus::Image::FromStream(logoStream, FALSE);
```
wcheng-msft: >g_image [](start = 7, length = 7)

appreciate removing unnecessary globals; please change the name though, thanks!

wcheng-msft: >		 [](start = 0, length = 2)

the project uses spaces instead of tabs.
for consistency, please also uses spaces

rudyhuyn: renamed!

rudyhuyn: I replaced all tabs in the file. I will push another PR to add a .editorconfig to the solution

PR #59, opened 2019-03-08
Changes to UI after app installation is complete

jyvenugo:
-Installed app launches if the 'Launch when ready' checkbox was checked and Msix app installer closes.
-If the checkbox is not checked, the launch window appears displaying the 'Launch' button.
-Clicking on Launch launches the app.
-Created InstallComplete handlers.
-Updated path so icon appears in control panel.
-Installed apps display only the publisher name now in 'Add or remove programs'.


preview/Win7Msix/Win7MSIXInstaller/InstallUI.hpp
```diff
@@ -118,3 +141,8 @@ BOOL ChangeText(HWND parentHWnd, std::wstring displayText, std::wstring  message
 //
 // PURPOSE: Increment the progress bar one tick based on preset tick
 void UpdateProgressBar();
+
+// FUNCTION: SendInstallCompleteMsg
+//
+// PURPOSE: Sends the WM_INSTALLCOMPLETE_MSG message to the main window when app installation is complete
+void SendInstallCompleteMsg();
```
wcheng-msft: >SendInstallCompleteMsg [](start = 5, length = 22)

can these be made members of the UI class? I hadn't gotten around to moving all the global stuff into the class, which is why there's so much stuff left as global. I'd rather not add more of this.

preview/Win7Msix/Win7MSIXInstaller/InstallUI.cpp
```diff
@@ -439,6 +484,24 @@ BOOL CreateCancelButton(HWND parentHWnd, RECT parentRect) {
 	return TRUE;
 }
 
+BOOL CreateLaunchButton(HWND parentHWnd, RECT parentRect) {
```
wcheng-msft: > { [](start = 57, length = 2)

style-wise let's be consistent on making this on a separate line by itself

preview/Win7Msix/Win7MSIXInstaller/InstallUI.hpp
```diff
@@ -25,22 +29,33 @@ class UI
...
+		m_launchAppEvent = CreateEvent(NULL, FALSE, FALSE, NULL);
+	}
     
 public:
     HRESULT DisplayPackageInfo(HWND hWnd, RECT windowRect, std::wstring& displayText, std::wstring& messageText);
     int CreateInitWindow(HINSTANCE hInstance, int nCmdShow, const std::wstring& windowClass, const std::wstring& title);
 
     void SetButtonClicked() { SetEvent(m_buttonClickedEvent); }
+    void SetLaunchButtonClicked() { SetEvent(m_launchAppEvent); }
+    HANDLE getLaunchButtonEvent()
```
wcheng-msft: >g [](start = 11, length = 1)

caps for method names?

preview/Win7Msix/Win7MSIXInstaller/InstallUI.hpp
```diff
@@ -25,22 +29,33 @@ class UI
...
+		m_buttonClickedEvent = CreateEvent(NULL, FALSE, FALSE, NULL);
+		m_launchAppEvent = CreateEvent(NULL, FALSE, FALSE, NULL);
+	}
     
 public:
     HRESULT DisplayPackageInfo(HWND hWnd, RECT windowRect, std::wstring& displayText, std::wstring& messageText);
     int CreateInitWindow(HINSTANCE hInstance, int nCmdShow, const std::wstring& windowClass, const std::wstring& title);
 
     void SetButtonClicked() { SetEvent(m_buttonClickedEvent); }
+    void SetLaunchButtonClicked() { SetEvent(m_launchAppEvent); }
```
wcheng-msft: >SetLaunchButtonClicked [](start = 9, length = 22)

i don't see this being called. did i miss it or is there a future task to call this?

jyvenugo: I created the event so the UI will wait without exiting while waiting for the launch button to be clicked at the same time using MsgWaitForMultipleObjects. The event is not being set anywhere.

preview/Win7Msix/Win7MSIXInstaller/InstallUI.hpp
jyvenugo: Moved UI controls and methods to members of UI class. Updated call locations.

preview/Win7Msix/Win7MSIXInstaller/InstallUI.cpp
jyvenugo: Moved bracket to new line.

preview/Win7Msix/Win7MSIXInstaller/InstallUI.hpp
jyvenugo: Updated method name.

preview/Win7Msix/Win7MSIXInstaller/InstallUI.hpp
wcheng-msft: how does it work currently if you aren't setting the event anywhere? i guess what i was asking was did you mean to call it somewhere and forget to add it? or are you planning on calling it somewhere in the future checkin as part of a different task? or are you finished with this and intend for it never to be set?

---
In reply to: [264356252](https://github.com/Microsoft/msix-packaging/pull/59#discussion_r264356252) [](ancestors = 264356252)

preview/Win7Msix/Win7MSIXInstaller/InstallUI.cpp
```diff
@@ -145,8 +145,37 @@ LRESULT CALLBACK WndProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
...
+        if (g_launchCheckBoxState) {
+            ui->LaunchInstalledApp(); // launch app
+            DestroyWindow(hWnd); // close msix app installer
+        }
+        else
+        {
+            //wait for user to click launch button or close the window
+            while (true)
+            {
+                MsgWaitForMultipleObjects(0, NULL, FALSE, INFINITE, QS_ALLINPUT);
```
wcheng-msft: >MsgWaitForMultipleObjects [](start = 16, length = 25)

even though you're waiting infinitely for 0 handles, you should still check the return value. It should only ever return one value, but you should still check it. Also the open brace (for the while) should be on a line by itself

PR #69, opened 2019-03-14
Split win7MsixInstaller in 2 parts: a standalone library and a .exe providing the UI.

### LINKED ISSUE #65: Win7MsixInstaller should be splitted into 2 parts: a library and a .exe managing the UI
In order to allow other developers and partners to use and integrate `Win7MsixInstaller` in their workflow/tools, the UI should be separated from the rest of the project.

**Win7MSIXInstaller Project** should only contains:
- Command line Parser
- Installer/Uninstaller UI

while the library **Win7MSIXInstallerLib** should manage the installation, uninstallation and information retrieval.

here are some examples of what we could do:
- provide a custom UI 
- use a different UI framework 
- manage the installation of a msix in an existing workflow without doing `Process.Start("win7msixInstaller.exe ....");`
- provide a different UX
- self-extracting installer for example
- ...

rudyhuyn:
## Fix #65

In order to allow other developers and partners to use and integrate Win7MsixInstaller in their workflow/tools, it's important to not only provide a .exe but also a library, easy to integrate in existing tools and significantly increasing the number of scenarios/use-cases possible with win7MSIX. The sooner this separation will be done, the easier it will be.

Here is how the split was done:

### Win7MSIXInstaller.exe
contains:
- Command line Parser
- Installer/Uninstaller UI

### static library Win7MSIXInstallerLib
Standalone library, managing all the existing features: installation, uninstallation, package listing, etc... but can be used by other applications, some examples:
- create an exe with a custom UI
- use a different UI framework
- manage the installation of a msix in an existing workflow without doing `Process.Start("win7msixInstaller.exe ....");`
- provide a different UX
- self-extracting installer for example
...

To validate this abstraction, we have created many applications using different UI tech that we can provide.

## Changes
All the current features are still present and works but all user interactions (win32 UI and std::cout) has been abstracted.

In order to isolate functions/classes from the library and not conflict with the client, a namespace has been added `Win7MsixInstallerLib` (temporary name).

The public and private header files of `Win7MsixInstallerLib` are separated in 2 directories:
- `Win7MsixInstallerLib/inc` the public headers.
- `Win7MsixInstallerLib/` the private headers.

To use the library:
```
IPackageManager ** packageManager;
Win7MsixInstallerLib_CreatePackageManager(&packageManager);
packageManager->AddPackage(filePath, options, [](status){ });
```
also support:
- Remove Package
- List Packages
- Find Package by Full name
- Find Package by Family name

Progress, Status and a callback has been added to the existing `MsixResponse` class in order to communicate with the UI.

## Interfaces 
Interfaces of this new library are based on existing Windows 10 Package Manager:

**Package/InstalledPackage** (based on https://docs.microsoft.com/en-us/uwp/api/windows.applicationmodel.package)
- Id
- DisplayName
- InstalledLocation
- Logo
- PublisherDisplayName
- Version

**interface IPackageManager**: (based on https://docs.microsoft.com/en-us/uwp/api/Windows.Management.Deployment.PackageManager)
- IMsixResponse AddPackage(String path, DeploymentOptions options, Func<IDeploymentResult> callback);  //uri should be file:/// but we can replace it by a filepath instead
- IPackage FindPackage(String packageFamilyName)
- IEnumerable<IPackage> FindPackages()
- IMsixResponse RemovePackage(String packageFullName, Func<IDeploymentResult> callback);
-IPackage GetPackageInfo(String path) <-- the only function not existing on Windows 10, but necessary to build the UI.

### Note:

The current architecture mixed UI and installation process, in fact, the UI was controlled/managed by the installation steps and fully part of the process, in the new architecture the UI is 100% independent and simply send command to the `PackageManager`.

I did my best to keep the current architecture (especially how `MsixRequest.cpp` works).

For info, the indentation of the namespaces in `Win7MsixInstallerLib` is not correct. 

```
namespace Win7MsixInstallerLib
{
class MsixRequest
{
```
instead of

```
namespace Win7MsixInstallerLib
{
    class MsixRequest
    {
```
It's on purpose, to simplify the readability of this diff. Similarly, this PR doesn't move files from Win7MsixInstaller to Win7MsixInstallerLib,  Git has some troubles to recognize that the file has been moved (and not deleted + added) if a lot of lines were modified.  

If this PR is accepted, we will move files and fix the formatting with another PR.

I'm available if you have feedback or questions!


preview/Win7Msix/Win7MSIXInstallerLib/inc/Win7MSIXInstallerActions.hpp
```diff
@@ -0,0 +1,10 @@
+#pragma once
+#include "DllExport.hpp"
+#include "IMsixRequest.hpp"
+#include <AppxPackaging.hpp>
+#include <vector>
+
+extern "C" DLLEXPORT HRESULT Win7MsixInstallerLib_CreateAddPackageRequest(std::wstring packageFilePath, MSIX_VALIDATION_OPTION validationOption, Win7MsixInstallerLib::IMsixRequest** outInstance);
```
sarjanas-msft: >Win7MsixInstallerLib_CreateAddPackageRequest [](start = 29, length = 44)

@rudyhuyn - Thanks for sharing this PR. I have sent you an email with some details about the API surface to be exposed from this lib that will meet more customer scenarios and requirements. Let's follow up in email and we can proceed accordingly with this PR. Thanks!

rudyhuyn: Done!

preview/Win7Msix/Win7MSIXInstallerLib/FilePaths.cpp
```diff
@@ -1,138 +1,151 @@
...
-}
+#include "FilePaths.hpp"
+#include "GeneralUtil.hpp"
+#include <shlobj_core.h>
+#include <KnownFolders.h>
+#include <mutex>
+
+using namespace Win7MsixInstallerLib;
+
+std::mutex filePathMappings_Init_lock;
```
wcheng-msft: how is this used?

preview/Win7Msix/Win7MSIXInstallerLib/inc/IPackageManager.hpp
```diff
@@ -0,0 +1,19 @@
...
+#include <functional>
+#include "IPackageInfo.hpp"
+#include "DeploymentOptions.hpp"
+#include "DeploymentResult.hpp"
+
+namespace Win7MsixInstallerLib {
+    class IPackageManager
+    {
+    public:
+        virtual bool AddPackage(const std::wstring & packageFilePath, DeploymentOptions options, std::function<void(const DeploymentResult &)> callback = nullptr) = 0;
```
wcheng-msft: >bool [](start = 16, length = 4)

why not just return the HRESULT?

preview/Win7Msix/Win7MSIXInstallerLib/PackageInfo.hpp
```diff
@@ -0,0 +1,114 @@
...
+    public:
+        PackageInfo() :PackageInfoBase() {}
+
+        std::wstring GetPackageFullName() { return m_packageFullName; }
+        std::wstring GetRelativeExecutableFilePath() { return m_relativeExecutableFilePath; }
+        std::wstring GetDisplayName() { return m_displayName; }
+        std::wstring GetId() { return m_appUserModelId; }
+        unsigned long long GetVersion() { return m_version; }
+        std::wstring GetPublisher() { return m_publisher; }
+        std::wstring GetPublisherDisplayName() { return m_publisherName; }
```
wcheng-msft: these are the same as the base class right?

preview/Win7Msix/Win7MSIXInstallerLib/MsixRequest.cpp
```diff
@@ -168,7 +127,7 @@ HRESULT MsixRequest::ProcessAddRequest()
         HandlerInfo currentHandler = AddHandlers[currentHandlerName];
         AutoPtr<IPackageHandler> handler;
         RETURN_IF_FAILED(currentHandler.create(this, &handler));
-        RETURN_IF_FAILED(handler->ExecuteForAddRequest());
+        RETURN_IF_FAILED(handler->ExecuteForAddRequest(packageInfo, installationPath));
```
wcheng-msft: >installationPath [](start = 68, length = 16)

if everything gets passed this, might as well make it part of the packageInfo?

preview/Win7Msix/Win7MSIXInstallerLib/PackageInfo.hpp
rudyhuyn: Yes, but we can't remove them, the interfaces IPackageInfo and IInstalledPackageInfo (Diamond interface inheritance problem)

preview/Win7Msix/Win7MSIXInstallerLib/inc/IPackageManager.hpp
rudyhuyn: HResult is very C++ oriented, I preferred to create a more generic interface and be closer to Win10 APIs

wcheng-msft: what Win10 APIs are you referring to?

---
In reply to: [270573968](https://github.com/Microsoft/msix-packaging/pull/69#discussion_r270573968) [](ancestors = 270573968)

preview/Win7Msix/Win7MSIXInstaller/MsixRequest.cpp
```diff
@@ -55,110 +53,71 @@ std::map<PCWSTR, HandlerInfo> AddHandlers =
...
-    return S_OK;
-}
+    if (packageInfo == nullptr)
+    {
+        return E_FAIL;
+    }
 
-HRESULT MsixRequest::ProcessAddRequest()
-{
-    PCWSTR currentHandlerName = PopulatePackageInfo::HandlerName;
```
wcheng-msft: >PopulatePackageInfo [](start = 32, length = 19)

we wanted not to have free-floating work done outside of the handlers, which is why we created a PopulatePackageInfo handler to encapsulate that work.

preview/Win7Msix/Win7MSIXInstallerLib/MsixRequest.cpp
```diff
@@ -55,110 +53,71 @@ std::map<PCWSTR, HandlerInfo> AddHandlers =
...
+    auto filemapping = FilePathMappings::GetInstance();
+    auto res = filemapping.GetInitializationResult();
+    if (FAILED(res))
     {
-        std::cout << p.path().filename() << std::endl;
-        numPackages++;
+        return E_FAIL;
     }
+    PackageInfo* packageInfo;
+    PopulatePackageInfo::GetPackageInfoFromPackage(this->m_packageFilePath.data(), this->m_validationOptions, &packageInfo);
```
wcheng-msft: >GetPackageInfoFromPackage [](start = 25, length = 25)

this can return errors, why convert every error into E_FAIL?

preview/Win7Msix/Win7MSIXInstallerLib/inc/IPackageManager.hpp
rudyhuyn: See the first post, the idea is to mimick existing Windows 10 WinRT API: Windows::ApplicationModel::PackageManager, Windows::ApplicationModel::Package, etc... and they never return a HRESULT. 

The error code (with errorMessage) will be added to `DeploymentResult` (I will do it today).

preview/Win7Msix/Win7MSIXInstallerLib/MsixRequest.cpp
rudyhuyn: I would prefer to keep `PackageInfo` limited to its purpose (info about a package), but we can create another class `RequestInfo` containing a packageInfo and the path (+ other info in the future)

preview/Win7Msix/Win7MSIXInstallerLib/FilePaths.cpp
rudyhuyn: No more used, I will remove it!

preview/Win7Msix/Win7MSIXInstaller/MsixRequest.cpp
rudyhuyn: In this case, we should also include `PopulatePackageInfo` in `RemoveHandlers (the current version on master calls it before the loop), what's your opinion?

rudyhuyn:
I took feedback into account and made 2 changed:
- 2 new classes added: `AddRequestInfo` and `RemoveRequestInfo` used by `ExecuteForAddRequest` and `ExecuteForRemoveRequest` to share information between handlers so we can put `PopulatePackageInfo` back in `AddHandlers`.
- `PopulatePackageInfo` was not part of `RemoveHandlers` (I suppose because this one was mandatory compared to others), so I added 2 methods `IPackageHandler::IsMandatoryForRemoveRequest()` and `IPackageHandler::IsMandatoryForAddRequest()` so we can ignore errors from a handler if not mandatory (all handlers except `PopulatePackageInfo` for RemovePackage)

rudyhuyn:
Hey everyone, I merge with master, simplify this diff to only focus on the split and take advantage of the recently added class `MsixResponse` (add `Percentage`, `Status` and `Cancel()`) so the .exe part will only need this object to adapt the UI. Let me know how we can move forward now!

preview/Win7Msix/Win7MSIXInstallerLib/inc/IPackageManager.hpp
```diff
@@ -0,0 +1,23 @@
...
+#include "IMsixResponse.hpp"
+#include "IPackage.hpp"
+#include "DeploymentOptions.hpp"
+
+
+namespace Win7MsixInstallerLib {
+    class IPackageManager
+    {
+    public:
+        virtual IMsixResponse * AddPackageAsync(const std::wstring & packageFilePath, DeploymentOptions options) = 0;
```
wcheng-msft: >AddPackageAsync [](start = 32, length = 15)

i think this might make more sense to take in the callback function, rather than expose SetCallback on the IMsixResponse interface; internally, we'd want to set the callback asap before returning the imsixresponse object because the addpackage will begin processing as the imsixresponse object is returned. 

Also, we need to make the msixresponse pointer a shared pointer so that it doesn't get deleted when msixrequest finishes.

rudyhuyn: What do you prefer:

```
IMsixResponse * response = CreateMsixResponse();
response.SetCallback(...);
HResult res = AddPackageAsync(packageFilePath, options, &response)
```

or 

```
HResult res = AddPackageAsync(packageFilePath, options, [](){ ... })
```

?

wcheng-msft: I think IMsixResponse * response = AddPackageAsync(packageFilePath, options, [](){...}); would be best. Not sure if you meant for the second of your options return value to change to HRESULT because now the response object is nowhere to be found?

rudyhuyn: Faulty copy/paste. I will modify that asap

rudyhuyn: @wcheng-msft, all pointers exposed by the library are now shared_ptr/unique_ptr + callback added

preview/Win7Msix/Win7MSIXInstaller/IPackageHandler.hpp
```diff
@@ -9,12 +12,14 @@ class IPackageHandler
     virtual HRESULT ExecuteForAddRequest() = 0;
 
     virtual HRESULT ExecuteForRemoveRequest() { return S_OK; }
+    virtual bool IsMandatoryForRemoveRequest() { return false; }
+    virtual bool IsMandatoryForAddRequest() { return true; }
```
sarjanas-msft: Why do we need these APIs?

rudyhuyn: For RemovePackage for example, some handlers are mandatory (PopulatePackageInfo) while others can be ignored if they return an error

preview/Win7Msix/Win7MSIXInstaller/MsixRequest.cpp
```diff
@@ -183,10 +130,24 @@ HRESULT MsixRequest::ProcessAddRequest()
...
-        HRESULT hrAddExecute = handler->ExecuteForAddRequest();
-        if (FAILED(hrAddExecute))
+        auto hrExecute = currentHandler.create(this, &handler);
+        if (FAILED(hrExecute))
+        {
+            if (handler->IsMandatoryForAddRequest())
+            {
+                m_msixResponse->SetHResultTextCode(hrExecute);
+                m_msixResponse->SetTextStatus(L"Can't create the handler " + std::wstring(currentHandlerName));
+                m_msixResponse->Update(InstallationStep::InstallationStepError, 0);
```
sarjanas-msft: Can we add this to a helper function?

preview/Win7Msix/Win7MSIXInstallerLib/inc/IPackageManager.hpp
```diff
@@ -0,0 +1,21 @@
...
+namespace Win7MsixInstallerLib {
+    class IPackageManager
+    {
+    public:
+        virtual IMsixResponse * AddPackage(const std::wstring & packageFilePath, DeploymentOptions options) = 0;
+        virtual IMsixResponse * RemovePackage(const std::wstring & packageFullName) = 0;
+        virtual IInstalledPackageInfo * FindPackage(const std::wstring & packageFamilyName) = 0;
+        virtual IInstalledPackageInfo * FindPackageByFamilyName(const std::wstring & packageFamilyName) = 0;
+        virtual std::vector<IInstalledPackageInfo*> * FindPackages() = 0;
+        virtual IPackage * GetPackageInfoMsix(const std::wstring & msixFullPath) = 0;
```
sarjanas-msft: GetMsixPackageInfo - suggested update to the name to be consistent

PR #71, opened 2019-03-20
Uninstall app when Cancel button is clicked

jyvenugo:
-When Cancel is clicked, a global variable in MsixRequest gets set to indicate that the install request was cancelled. The rest of the handlers skip execution and a ProcessRemoveRequest() is trigerred at the end of the handler chain if this variable was set.
-Removed g_installed global variable from InstallUI.cpp as this was not being set anywhere.
-Updated the output text of FindAllPackages command to include a bracket for package(s) found.
-Moved comments in the file InstallUI.cpp to the header file.


preview/Win7Msix/Win7MSIXInstaller/Extractor.cpp
```diff
@@ -276,9 +276,12 @@ HRESULT Extractor::CreateHandler(MsixRequest * msixRequest, IPackageHandler ** i
 
 HRESULT Extractor::ExtractPackage()
 {
-    RETURN_IF_FAILED(ExtractFootprintFiles());
-    RETURN_IF_FAILED(ExtractPayloadFiles());
-    RETURN_IF_FAILED(ExtractRegistry(false));
+    if(!m_msixRequest->GetIsCancelClicked())
+        RETURN_IF_FAILED(ExtractFootprintFiles());
+    if (!m_msixRequest->GetIsCancelClicked())
+        RETURN_IF_FAILED(ExtractPayloadFiles());
```
wcheng-msft: >ExtractPayloadFiles [](start = 25, length = 19)

extractPayloadFiles should check between files as well, as that's generally going to be the most cancellable period of time.

Also, always use curly braces {} even for one line if statements, to avoid mistakes.

preview/Win7Msix/Win7MSIXInstaller/MsixRequest.cpp
```diff
@@ -218,4 +227,19 @@ void MsixRequest::SetUI(UI * ui)
...
+    RETURN_IF_FAILED(MsixRequest::Make(OperationType::Remove, Flags::NoFlags, std::wstring(), packageFullName, MSIX_VALIDATION_OPTION::MSIX_VALIDATION_OPTION_FULL, &removePackageRequest));
+
+    const HRESULT hrCancelRequest = removePackageRequest->ProcessRequest();
+    if (FAILED(hrCancelRequest))
+    {
+        TraceLoggingWrite(g_MsixTraceLoggingProvider,
+            "Failed to process cancel request",
+            TraceLoggingLevel(WINEVENT_LEVEL_WARNING),
+            TraceLoggingValue(hrCancelRequest, "HR"));
+    }
```
wcheng-msft: you should be getting a compiler warning or error that there's a codepath that doesn't return an HRESULT? should be return S_OK at the end.

preview/Win7Msix/Win7MSIXInstaller/MsixRequest.hpp
```diff
@@ -62,6 +66,9 @@ class MsixRequest
     /// @return can return null if called before CreateAndShowUI or if Flags::QuietUX was passed in and there is no UI.
     UI* GetUI() { return m_UI; }
     bool IsQuietUX() { return (m_flags & Flags::QuietUX) == Flags::QuietUX; }
+
+    /// Prepares and sends msixrequest to uninstall the package in case cancel button was clicked during installation
+    HRESULT UnInstallPackage(std::wstring packageFullName);
```
wcheng-msft: >UnInstallPackage [](start = 12, length = 16)

prefer to call this RemovePackage

preview/Win7Msix/Win7MSIXInstaller/MsixRequest.hpp
```diff
@@ -74,6 +81,17 @@ class MsixRequest
         return true;
     }
 
+    bool GetIsCancelClicked()
```
wcheng-msft: >GetIsCancelClicked [](start = 9, length = 18)

Just call it GetIsCancelled or GetIsInstallCancelled to be consistent with the variable name m_isInstallCancelled;

preview/Win7Msix/Win7MSIXInstaller/MsixRequest.hpp
```diff
@@ -37,6 +37,9 @@ class MsixRequest
     /// Filled in by CreateAndShowUI 
     AutoPtr<UI> m_UI;
 
+    ///Variable used to indicate if add package request was cancelled during installation
+    bool m_isInstallCancelled;
```
wcheng-msft: >m_isInstallCancelled [](start = 9, length = 20)

initialize = false

preview/Win7Msix/Win7MSIXInstaller/MsixRequest.hpp
```diff
@@ -74,6 +81,17 @@ class MsixRequest
         return true;
     }
 
+    bool GetIsCancelClicked()
+    {
+        return m_isInstallCancelled;
+    }
+
+    //Setters
+    void SetIsCancelClicked()
```
wcheng-msft: >SetIsCancelClicked [](start = 9, length = 18)

could just be Cancel() or CancelRequest() as a verb.

preview/Win7Msix/Win7MSIXInstaller/InstallUI.hpp
```diff
@@ -20,14 +20,13 @@ static HWND g_checkboxHWnd = NULL;
 static HWND g_progressHWnd = NULL;
 static HWND g_CancelbuttonHWnd = NULL;
 static HWND g_LaunchbuttonHWnd = NULL;
-static bool g_installed = false;
 static bool g_launchCheckBoxState = true; // launch checkbox is checked by default
 
 class UI
```
sarjanas-msft: >UI [](start = 6, length = 2)

I would recommend making this more descriptive name

preview/Win7Msix/Win7MSIXInstaller/MsixRequest.cpp
jyvenugo: Modified to return S_OK.

preview/Win7Msix/Win7MSIXInstaller/MsixRequest.hpp
jyvenugo: Modified method name to RemovePackage.

sarjanas-msft:
>     // count: the number of objects to be iterated through in the progress bar

Lets follow consistent commenting practice. I believe William has already used doxygen style comments with /// for instance. 

---
Refers to: preview/Win7Msix/Win7MSIXInstaller/InstallUI.hpp:80 in 6c7c566. [](commit_id = 6c7c5665aae2b67f9f3671ec575773ea0c3b0896, deletion_comment = False)

preview/Win7Msix/Win7MSIXInstaller/MsixRequest.hpp
jyvenugo: Modified method name to GetIsInstallCancelled().

preview/Win7Msix/Win7MSIXInstaller/MsixRequest.hpp
jyvenugo: Modified.

preview/Win7Msix/Win7MSIXInstaller/MsixRequest.hpp
jyvenugo: Modified method name to CancelRequest().

preview/Win7Msix/Win7MSIXInstaller/Extractor.cpp
jyvenugo: Added check for cancel button within extractPayloadFiles as well. Added braces.

preview/Win7Msix/Win7MSIXInstaller/MsixRequest.cpp
```diff
@@ -168,11 +168,20 @@ HRESULT MsixRequest::ProcessAddRequest()
         HandlerInfo currentHandler = AddHandlers[currentHandlerName];
         AutoPtr<IPackageHandler> handler;
         RETURN_IF_FAILED(currentHandler.create(this, &handler));
-        RETURN_IF_FAILED(handler->ExecuteForAddRequest());
+        if (!GetIsCancelClicked())
```
sarjanas-msft: >GetIsCancelClicked [](start = 13, length = 18)

Shouldn't this be checked in conjunction with the show UI option? Also, I do not believe we should add any code to this loop. If we need to handle a remove, we should do that explicitly by creating a remove request, or optionally in the array structure above, we can go from one statehandler to another in cancel case, and that can be the cancel handler.

preview/Win7Msix/Win7MSIXInstaller/MsixRequest.hpp
sarjanas-msft: >UnInstallPackage [](start = 12, length = 16)

If we want a first class notion of cancel we have to figure out a different way to expose it. It should not be an API on the request.

preview/Win7Msix/Win7MSIXInstaller/AddRemovePrograms.cpp
```diff
@@ -13,6 +13,11 @@ const PCWSTR AddRemovePrograms::HandlerName = L"AddRemovePrograms";
 
 HRESULT AddRemovePrograms::ExecuteForAddRequest()
 {
+    if (m_msixRequest->GetIsInstallCancelled())
+    {
+        return ERROR_INSTALL_USEREXIT;
```
wcheng-msft: >ERROR_INSTALL_USEREXIT [](start = 15, length = 22)

this is win32, you need to HRESULT_FROM_WIN32 to return an HRESULT.

preview/Win7Msix/Win7MSIXInstaller/MsixRequest.cpp
```diff
@@ -168,9 +174,17 @@ HRESULT MsixRequest::ProcessAddRequest()
         HandlerInfo currentHandler = AddHandlers[currentHandlerName];
         AutoPtr<IPackageHandler> handler;
         RETURN_IF_FAILED(currentHandler.create(this, &handler));
-        RETURN_IF_FAILED(handler->ExecuteForAddRequest());
-
-        currentHandlerName = currentHandler.nextHandler;
+        HRESULT hr = handler->ExecuteForAddRequest();
+        if (FAILED(hr))
+        {
+            m_msixResponse->SetErrorCode(hr);
```
wcheng-msft: >m_msixResponse [](start = 12, length = 14)

i think when we talked we wanted to pass the response object to the handlers, and have the response object track that it's canceled instead of the msixRequest., and that we'd call a special error handler which is tracked in the HandlerInfo above.

preview/Win7Msix/Win7MSIXInstaller/AddRemovePrograms.cpp
jyvenugo: Updated to return HRESULT_FROM_WIN32

preview/Win7Msix/Win7MSIXInstaller/MsixRequest.cpp
```diff
@@ -26,30 +26,37 @@
...
+{
+    CreateHandler create;
+    PCWSTR handlerName;
+};
+
 struct HandlerInfo
 {
     CreateHandler create;
     PCWSTR nextHandler;
+    ErrorHandlerInfo errorHandler;
```
wcheng-msft: >ErrorHandlerInfo [](start = 4, length = 16)

this could be simpler if this was just PCWSTR for the name, and that the errorHandler was just one of the HandlerInfo in the array instead of a separately typed ErrorHandlerInfo

preview/Win7Msix/Win7MSIXInstaller/MsixRequest.hpp
```diff
@@ -50,10 +54,13 @@ class MsixRequest
     /// Called by CreateAndShowUI 
     void SetUI(UI* ui);
 
+    void SetMsixResponse(MsixResponse* msixResponse);
```
wcheng-msft: >SetMsixResponse [](start = 9, length = 15)

is this being called outside? by whom?  my thoughts are either 1. this is created by caller and passed in through MsixRequest::Make(), or 2. the MsixRequest::Make itself creates it like you're doing but then doesn't expose a way to set it

preview/Win7Msix/Win7MSIXInstaller/FileTypeAssociation.cpp
```diff
@@ -142,11 +142,21 @@ HRESULT FileTypeAssociation::ParseManifest()
...
+
     for (std::vector<Fta>::iterator fta = m_Ftas.begin(); fta != m_Ftas.end(); ++fta)
     {
         RETURN_IF_FAILED(ProcessFtaForAdd(*fta));
     }
 
+    if (m_msixRequest->GetMsixResponse()->GetIsInstallCancelled())
+    {
+        return HRESULT_FROM_WIN32(ERROR_INSTALL_USEREXIT);
+    }
```
wcheng-msft: i don't think we need this to be checked before AND after each handler because there is hardly any work between handlers and we expect each handler to be checking it. However, it might make sense to do it after each FTA (i.e. inside the loop) in case each FTA has significant work.

preview/Win7Msix/Win7MSIXInstaller/MsixRequest.cpp
```diff
@@ -77,6 +84,14 @@ HRESULT MsixRequest::Make(OperationType operationType, Flags flags, std::wstring
     instance->m_packageFullName = packageFullName;
     instance->m_validationOptions = validationOption;
     RETURN_IF_FAILED(instance->InitializeFilePathMappings());
+
+    //Set MsixResponse
+    AutoPtr<MsixResponse> localResponse;
+    RETURN_IF_FAILED(MsixResponse::Make(
+        &localResponse)
+    );
```
wcheng-msft: prefer to make this one line

preview/Win7Msix/Win7MSIXInstaller/InstallComplete.hpp
```diff
@@ -6,14 +6,15 @@
 class InstallComplete : IPackageHandler
 {
 public:
-	HRESULT ExecuteForAddRequest();
+    /// Send install complete message to UI if cancel button was not clicked, else uninstalls the app
```
wcheng-msft: > if cancel button was not clicked, else uninstalls the app [](start = 43, length = 58)

you're not uninstalling here, and i think at a certain point (this point) you reach the point of no return and can no longer cancel anyway, so i don't think you should mention anything about cancel.

preview/Win7Msix/Win7MSIXInstaller/MsixRequest.cpp
jyvenugo: Updated structure to include an entry for error handler.

preview/Win7Msix/Win7MSIXInstaller/MsixRequest.hpp
jyvenugo: It is not being called anywhere, I forgot to remove it, removed the same.

preview/Win7Msix/Win7MSIXInstaller/MsixRequest.cpp
jyvenugo: Updated.

preview/Win7Msix/Win7MSIXInstaller/InstallComplete.hpp
jyvenugo: Updated comment.

preview/Win7Msix/Win7MSIXInstaller/FileTypeAssociation.cpp
jyvenugo: Updated check for FTA handler and added check inside loop. Also, updated the same for Protocol, StartMenulink and AddRemovePrograms.

preview/Win7Msix/Win7MSIXInstaller/MsixRequest.cpp
jyvenugo: Modified the code as per design discussion to have an error handler which will remove the package in case of other handlers throw an error.

preview/Win7Msix/Win7MSIXInstaller/MsixRequest.hpp
jyvenugo: Modified and moved to error handler as per design discussion.

preview/Win7Msix/Win7MSIXInstaller/MsixResponse.hpp
```diff
@@ -0,0 +1,52 @@
...
+    {
+        errorCode = code;
+    }
+
+    HRESULT GetErrorCode()
+    {
+        return errorCode;
+    }
+
+    void setErrorText(PCWSTR errorText)
```
sarjanas-msft: >setErrorText [](start = 9, length = 12)

Fix the name casing

preview/Win7Msix/Win7MSIXInstaller/MsixResponse.hpp
```diff
@@ -0,0 +1,52 @@
...
+    {
+        errorText = errorText;
+    }
+
+    PCWSTR getErrorText()
```
sarjanas-msft: >getErrorText [](start = 11, length = 12)

Fix the name casing

preview/Win7Msix/Win7MSIXInstaller/MsixResponse.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+    HRESULT errorCode;
+
+    PCWSTR errorText;
+
+public:
+
+    static HRESULT Make(MsixResponse** outInstance);
+
+    //Setters and Getters
```
sarjanas-msft: >//Setters and Getters [](start = 4, length = 21)

Follow standard comments like we discussed with /// and more information about parameters, return values, potential errors etc.

preview/Win7Msix/Win7MSIXInstaller/MsixResponse.hpp
```diff
@@ -0,0 +1,52 @@
...
+    bool GetIsInstallCancelled()
+    {
+        return m_isInstallCancelled;
+    }
+
+    void CancelRequest()
```
sarjanas-msft: inline?

preview/Win7Msix/Win7MSIXInstaller/MsixResponse.hpp
```diff
@@ -0,0 +1,52 @@
+#pragma once
+#include <windows.h>
+#include <string>
+#include "GeneralUtil.hpp"
+
+class MsixResponse
```
sarjanas-msft: Wouldn't progress also go on this interface? We should add comments here to then as we merge with Rudy's change we can make sure it is aligned.

preview/Win7Msix/Win7MSIXInstaller/MsixResponse.hpp
jyvenugo: Updated.

preview/Win7Msix/Win7MSIXInstaller/MsixResponse.hpp
jyvenugo: Updated.

preview/Win7Msix/Win7MSIXInstaller/MsixResponse.hpp
jyvenugo: Added comments.

preview/Win7Msix/Win7MSIXInstaller/MsixResponse.hpp
jyvenugo: Modified getters and setters to inline.

preview/Win7Msix/Win7MSIXInstaller/MsixResponse.hpp
jyvenugo: Yes, added comments for progress bar changes.

preview/Win7Msix/Win7MSIXInstaller/Extractor.cpp
```diff
@@ -276,8 +280,22 @@ HRESULT Extractor::CreateHandler(MsixRequest * msixRequest, IPackageHandler ** i
...
+    if (m_msixRequest->GetMsixResponse()->GetIsInstallCancelled())
+    {
+        return HRESULT_FROM_WIN32(ERROR_INSTALL_USEREXIT);
+    }
     RETURN_IF_FAILED(ExtractPayloadFiles());
+
+    if (m_msixRequest->GetMsixResponse()->GetIsInstallCancelled())
+    {
+        return HRESULT_FROM_WIN32(ERROR_INSTALL_USEREXIT);
+    }
```
sarjanas-msft: Do we need these three if we are already checking in the loops within the methods called? This pattern does not look right to me and also makes the code less readable.

preview/Win7Msix/Win7MSIXInstaller/MsixRequest.hpp
```diff
@@ -74,6 +79,8 @@ class MsixRequest
         return true;
     }
```
sarjanas-msft: make these also inline

jyvenugo: Added inline to methods executing minimal code.

preview/Win7Msix/Win7MSIXInstaller/Extractor.cpp
jyvenugo: I added this since Cancel could be clicked at any time, it's not needed at all places. Moved it into the loop.

PR #80, opened 2019-04-05
Button and Initial UI screen should show Update if V1 is installed

jyvenugo:
This PR updates the button and initial install screen UI text to 'Update' if V1 is already installed on the machine. 
- Added method CheckIfUpdate() in InstallUI to check if V1 is already installed before displaying the UI.
- The same check occurs in ProcessPotentialHandler.cpp(which occurs after the user has clicked on Install/Update). I considered just doing it once, but that would require also storing the V1 packagename in msixrequest.cpp object so that ProcessPotentialRequest handler would know which package to remove.


preview/Win7Msix/Win7MSIXInstaller/InstallUI.hpp
```diff
@@ -22,6 +22,7 @@ static HWND g_CancelbuttonHWnd = NULL;
 static HWND g_LaunchbuttonHWnd = NULL;
 static bool g_launchCheckBoxState = true; /// launch checkbox is checked by default
 static bool g_installed = false;
+static PCWSTR g_installOrUpdateText = L"Install "; /// Default button and install screen UI text is 'Install'
```
wcheng-msft: >Install [](start = 40, length = 7)

we should be moving these UX elements over to string resources so that they can eventually be localized

wcheng-msft: >g_installOrUpdateText [](start = 14, length = 21)

i'd prefer this to be part of the class rather than global

jyvenugo: Moved to UI class. Called checking for an update after SetWindowLongPtr() and called ChangeButtonText later to set button text to 'Update' if the operation is an update.

jyvenugo: Moved strings to resource file.

preview/Win7Msix/Win7MSIXInstaller/InstallUI.hpp
```diff
@@ -28,13 +28,15 @@ class UI
 public:
     HRESULT ShowUI();
     HRESULT LaunchInstalledApp();
-    void ConfirmAppCancel();
+    void ConfirmAppCancel(HWND parentHWnd);
+    std::wstring g_installOrUpdateText = GetStringResource(IDS_STRING_INSTALLTEXT); /// Default button and install screen UI text is 'Install'
```
wcheng-msft: >g_installOrUpdateText [](start = 17, length = 21)

does this need to be public? also non-global shouldn't start with g_

preview/Win7Msix/Win7MSIXInstaller/InstallUI.hpp
```diff
@@ -53,6 +55,22 @@ class UI
 
 public:
 
+    /// This function sets the parent window hwnd after create window
+    ///
+    /// @param hWnd - the HWND of the parent window
+    void SetHwnd(HWND hwnd)
+    {
+        hWnd = hwnd;
+    }
```
wcheng-msft: does this need to be public?

wcheng-msft:
>     IDS_STRING_INSTALLTEXT  "Install "

i'd prefer the space to be added separately when needed and that the text just be text without trailing whitespace

---
Refers to: preview/Win7Msix/Win7MSIXInstaller/Win7MSIXInstaller.rc:175 in 8948f16. [](commit_id = 8948f16b160ae58e32e0154107bcd0b20bedd7ee, deletion_comment = False)

preview/Win7Msix/Win7MSIXInstaller/InstallUI.hpp
jyvenugo: No, this need not necessarily be public for now. Moved to private.

preview/Win7Msix/Win7MSIXInstaller/InstallUI.hpp
jyvenugo: Updated.

preview/Win7Msix/Win7MSIXInstaller/InstallUI.hpp
```diff
@@ -28,13 +28,15 @@ class UI
...
-    void ConfirmAppCancel();
+    void ConfirmAppCancel(HWND parentHWnd);
 
     static HRESULT Make(_In_ MsixRequest* msixRequest, _Out_ UI** instance);
     ~UI() {}
 private:
     MsixRequest* m_msixRequest = nullptr;
 
+    HWND hWnd = NULL; //Parent Window Hwnd
+    std::wstring installOrUpdateText = GetStringResource(IDS_STRING_INSTALLTEXT); /// Default button and install screen UI text is 'Install'
```
wcheng-msft: >installOrUpdateText [](start = 17, length = 19)

sorry, i should have mentioned, the convention we're using is to prefix member variables with m_ (see lines immediately below)

jyvenugo: Resolved.

PR #81, opened 2019-04-05
Allow explicit selection of crypto library

JohnMcPMS:
To allow testing of the OpenSSL implementation on Windows, this change creates a new CRYPTO_LIB setting in cmake.  The previous implicit selection is preserved through detection of the wincrypt.h header, and only makewin.cmd is updated to allow overriding with OpenSSL.

Additionally, the Win32.ps1 test script has various fixes and improvements to make the output more visually parsable (ie, colors).  It also collects the failing test 'names' and prints them at the end.

All existing tests in Win32.ps1 work when OpenSSL is selected, in both x86 and x64 builds.


src/msix/CMakeLists.txt
```diff
@@ -43,10 +43,28 @@ if(XML_PARSER MATCHES msxml6)
     add_definitions(-DUSING_MSXML=1)
 endif()
 
-if(WIN32)
-    set(DirectoryObject PAL/FileSystem/Win32/DirectoryObject.cpp)
+if(CRYPTO_LIB MATCHES crypt32)
```
msftrubengu: I don't except this to happen but we should probably fail here if someone doesn't use our builds scripts and do for example cmake -DLINUX=on -DCRYPTO_LIB=crypt32 ..

JohnMcPMS: I didn't see a similar error for XML_PARSER.  I can look into one for both, but ultimately the compile will just fail to find includes quite quickly.

msftrubengu: I am ok failing at compile time. The erros is going to be pretty obvious for people to figure out.

PR #82, opened 2019-04-09
Add 19H1 schemas and enable manifest schema validation for Xerces

msftrubengu:
- 19H1 schemas added/modified.
- Enable schema validation for Xerces. Some manifest had to be modified to accomplish this. This is because Xerces maxOccurs element limitations and non standard xs:pattern in the schemas .See msix_resources.cmake comments for more
- Implement removal of elements and attributes in a namespace for Xerces, where the namespace is in the top level IgnorableNamespaces attribute and we don't know about. 
- Add automated test for Linux for validation parser = on. This should be enough as Xerces is pretty reliable on different platforms. Currently we only test using non-validation Xerces for Windows and Linux.

TODOs:
- Do semantic validations to validate the changes we did to the schemas.
- Now that we have the packing prototype, we can recreate the negative win8 tests and create a win10 msix package that is invalid.


resources/AppxPackaging/Manifest/Schema/Xerces/AppxManifestTypes.xsd
```diff
@@ -0,0 +1,1326 @@
...
+
+@@END_SDKSPLIT -->
+<!--
+  Copyright (c) Microsoft Corporation. All rights reserved.
+
+  Defines common types shared by multiple namespaces.  This schema is imported
+  into various product-specific schemas that define elements and attributes
+  based on types declared here.
+-->
+<xs:schema attributeFormDefault="unqualified"
```
JohnMcPMS: What is this file for?

src/msix/PAL/XML/xerces-c/XmlObject.cpp
```diff
@@ -40,11 +45,81 @@ class IXercesElement : public IUnknown
 {
 public:
     virtual DOMElement* GetElement() = 0;
+    virtual std::string GetAttributeValue(std::string& attributeName) = 0;
```
JohnMcPMS: >std::string& [](start = 42, length = 12)

const

src/msix/PAL/XML/xerces-c/XmlObject.cpp
```diff
@@ -53,20 +128,52 @@ class ParsingException final : public XERCES_CPP_NAMESPACE::ErrorHandler
...
 
     void resetErrors() override {}
+private:
+    std::string GetMessage(const XERCES_CPP_NAMESPACE::SAXParseException& exp)
+    {
+        std::u16string utf16FileName = std::u16string(exp.getSystemId());
+        std::u16string utf16Message = std::u16string(exp.getMessage());
+        return "Error " + u16string_to_utf8(utf16Message) + " file: " + u16string_to_utf8(utf16FileName) +
+            ":" + std::to_string(static_cast<std::uint64_t>(exp.getLineNumber())) + "," +
+            std::to_string(static_cast<std::uint64_t>(exp.getColumnNumber()));
```
JohnMcPMS: Can we do something more like:

Error in FILENAME.XML [Line 1, Col 1] :: THIS IS AN ERROR MESSAGE

src/msix/PAL/XML/xerces-c/XmlObject.cpp
```diff
@@ -53,20 +128,52 @@ class ParsingException final : public XERCES_CPP_NAMESPACE::ErrorHandler
...
+    MsixEntityResolver(IMsixFactory* factory, const NamespaceManager& namespaces) : m_factory(factory), m_namespaces(namespaces) {}
+    ~MsixEntityResolver() {}
+
+    InputSource* resolveEntity(XMLResourceIdentifier* resourceIdentifier)
+    {
+        std::u16string utf16string = std::u16string(resourceIdentifier->getNameSpace());
+        std::string id = u16string_to_utf8(utf16string);
+        const auto& entry = std::find(m_namespaces.begin(), m_namespaces.end(), id.c_str());
+        ThrowErrorIf(MSIX::Error::XmlError, entry == m_namespaces.end(), "Invalid namespace");
+        auto stream = m_factory->GetResource((*entry).schema);
```
JohnMcPMS: >(*entry). [](start = 45, length = 9)

entry->    ?

src/msix/PAL/XML/xerces-c/XmlObject.cpp
```diff
@@ -402,6 +533,124 @@ class XercesDom final : public ComClass<XercesDom, IXmlDom>
...
+        static const XMLCh cs[3] = {chLatin_L, chLatin_S, chNull};
+        DOMImplementation *impl = DOMImplementationRegistry::getDOMImplementation(cs);
+        auto serializer = XercesPtr<DOMLSSerializer>((static_cast<DOMImplementationLS*>(impl))->createLSSerializer());
+        auto lsOutput = XercesPtr<DOMLSOutput>((static_cast<DOMImplementationLS*>(impl))->createLSOutput());
+
+        // Set encoding to UTF-8
+        static const XMLCh utf8Str[] = {chLatin_U, chLatin_T, chLatin_F, chDash, chDigit_8, chNull};
+        lsOutput->setEncoding(utf8Str);
+
+        // Don't use unique_ptr here, this buffer is going to be deleted by Xerces
```
JohnMcPMS: >Don't use unique_ptr here, this buffer is going to be deleted by Xerces [](start = 11, length = 71)

Comment is confusing since the very next thing you do is use a unique_ptr

src/inc/StreamHelper.hpp
```diff
@@ -28,5 +30,24 @@ namespace MSIX {
             ThrowHrIfFailed(stream->Seek(start, StreamBase::Reference::START, nullptr));
             return buffer;
         }
+
+        inline std::pair<std::uint32_t, std::uint8_t*> CreateRawBufferFromStream(const ComPtr<IStream>& stream)
```
JohnMcPMS: >std::pair<std::uint32_t, std::uint8_t*> [](start = 15, length = 39)

Are you shying away from a std::vector for a reason?  Does it bloat the binary?

src/msix/PAL/XML/xerces-c/XmlObject.cpp
```diff
@@ -402,6 +533,124 @@ class XercesDom final : public ComClass<XercesDom, IXmlDom>
     }
 
 protected:
+
+    std::unique_ptr<MemBufInputSource> StripIgnorableNamespaces(const InputSource& source, const NamespaceManager& namespaces)
```
JohnMcPMS: >StripIgnorableNamespaces [](start = 39, length = 24)

Would be nice if this could live above the PAL, but the cost of projecting everything that would be needed seems high.

resources/AppxPackaging/Manifest/Schema/Xerces/AppxManifestTypes.xsd
msftrubengu: Not needed, I forgot to delete it...

---
In reply to: [273616604](https://github.com/Microsoft/msix-packaging/pull/82#discussion_r273616604) [](ancestors = 273616604)

src/msix/PAL/XML/xerces-c/XmlObject.cpp
msftrubengu: yeah, this comment was supposed to be before copying the buffer.  moved

---
In reply to: [273624046](https://github.com/Microsoft/msix-packaging/pull/82#discussion_r273624046) [](ancestors = 273624046)

src/inc/StreamHelper.hpp
msftrubengu: talked offline, this will change to std::pair<std::uint32_t, std::unique_ptr<std::uint8_t>>

---
In reply to: [273627988](https://github.com/Microsoft/msix-packaging/pull/82#discussion_r273627988) [](ancestors = 273627988)

src/inc/MSIXResource.hpp.cmakein
```diff
@@ -0,0 +1,56 @@
+//
+//  Copyright (C) 2019 Microsoft.  All rights reserved.
+//  See LICENSE file in the project root for full license information.
+//
+//  This file is generated by CMake. Do not edit.
```
JohnMcPMS: This isn't quite true any more.

src/inc/MSIXResource.hpp.cmakein
```diff
@@ -0,0 +1,56 @@
...
+//
+#include "AppxPackaging.hpp"
+#include "ComHelper.hpp"
+#include "AppxFactory.hpp"
+#include <map>
+
+namespace MSIX {
+
+    ${NAMESPACE_MANAGER}
```
JohnMcPMS: It feels like this should get dumped into its own cpp with an extern reference here.

src/inc/MSIXResource.hpp.cmakein
msftrubengu: This is still true for the header. CMake takes MSIXResource.hpp.cmakein and creates MSIXResource.hpp 

---
In reply to: [274102137](https://github.com/Microsoft/msix-packaging/pull/82#discussion_r274102137) [](ancestors = 274102137)

JohnMcPMS: It makes the hpp, but the cmakein file (this one) isn't generated and it does need to be modified potentially.

---
In reply to: [274121009](https://github.com/Microsoft/msix-packaging/pull/82#discussion_r274121009) [](ancestors = 274121009,274102137)

msftrubengu: Ok, modified to "MSIXResource.hpp is generated by CMake. Do not edit"

---
In reply to: [274121740](https://github.com/Microsoft/msix-packaging/pull/82#discussion_r274121740) [](ancestors = 274121740,274121009,274102137)

PR #87, opened 2019-04-12
Startuptask

wcheng-msft:
Added startup task handling -- this uses TaskSchd APIs to create a scheduled task to run on logon. We don't have OS support to support the app itself being able to enable/disable/query anything with Windows.ApplicationModel.StartupTask, so that aspect is omitted. I looked at using the Run key, but that apparently only works when UAC is disabled even for exes that do not require UAC.

Fixed a few crash scenarios.


preview/Win7Msix/Win7MSIXInstaller/MsixRequest.cpp
```diff
@@ -46,15 +47,16 @@ struct HandlerInfo
 std::map<PCWSTR, HandlerInfo> AddHandlers =
 {
     //HandlerName                         Function to create                      NextHandler                          ErrorHandlerInfo
-    {PopulatePackageInfo::HandlerName,    {PopulatePackageInfo::CreateHandler,    CreateAndShowUI::HandlerName,        ErrorHandler::HandlerName}},
-    {CreateAndShowUI::HandlerName,        {CreateAndShowUI::CreateHandler,        ProcessPotentialUpdate::HandlerName, ErrorHandler::HandlerName}},
+    {PopulatePackageInfo::HandlerName,    {PopulatePackageInfo::CreateHandler,    CreateAndShowUI::HandlerName,        nullptr}},
```
sarjanas-msft: why is this now nullptr? It might be good to put inline comments to clarify.

preview/Win7Msix/Win7MSIXInstaller/StartupTask.cpp
```diff
@@ -0,0 +1,193 @@
...
+const PCWSTR StartupTask::TaskDefinitionXmlPostfix =
+                L"\"</Command>\
+            </Exec>\
+        </Actions>\
+    </Task>";
+
+const PCWSTR windowsTaskFolderName = L"\\Microsoft\\Windows";
+const PCWSTR taskFolderName = L"MsixCore";
+
+HRESULT StartupTask::ExecuteForAddRequest()
```
sarjanas-msft: Please add comments to the methods to briefly describe the algorithm used to add the task and similarly for the remove case. Also document any known errors that the caller can expect.

preview/Win7Msix/Win7MSIXInstaller/StartupTask.cpp
```diff
@@ -0,0 +1,193 @@
...
+#include "FilePaths.hpp"
+#include "StartupTask.hpp"
+#include "GeneralUtil.hpp"
+#include "Constants.hpp"
+#include <TraceLoggingProvider.h>
+#include <taskschd.h>
+
+const PCWSTR StartupTask::HandlerName = L"StartupTask";
+
+const PCWSTR StartupTask::TaskDefinitionXmlPrefix =
```
sarjanas-msft: I am assuming that this definition is compatible with what ships in Win10. May be add a comment to that effect.  Also are there any settings that can be influenced based on the xml in the manifest?

preview/Win7Msix/Win7MSIXInstaller/StartupTask.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+const PCWSTR windowsTaskFolderName = L"\\Microsoft\\Windows";
+const PCWSTR taskFolderName = L"MsixCore";
+
+HRESULT StartupTask::ExecuteForAddRequest()
+{
+    if (m_executable.empty())
+    {
+        return S_OK;
+    }
```
sarjanas-msft: What is the scenario for this? Add comments to clarify.

preview/Win7Msix/Win7MSIXInstaller/StartupTask.cpp
```diff
@@ -0,0 +1,193 @@
...
+        ComPtr<IMsixElement> extensionElement;
+        RETURN_IF_FAILED(extensionEnum->GetCurrent(&extensionElement));
+        Text<wchar_t> extensionCategory;
+        RETURN_IF_FAILED(extensionElement->GetAttributeValue(categoryAttribute.c_str(), &extensionCategory));
+
+        if (wcscmp(extensionCategory.Get(), startupTaskCategoryNameInManifest.c_str()) == 0)
+        {
+            Text<wchar_t> executable;
+            RETURN_IF_FAILED(extensionElement->GetAttributeValue(executableAttribute.c_str(), &executable));
+            m_executable = m_msixRequest->GetFilePathMappings()->GetExecutablePath(executable.Get(), m_msixRequest->GetPackageInfo()->GetPackageFullName().c_str());
```
sarjanas-msft: Does schema validation guarantee that there is exactly one of these defined?

preview/Win7Msix/Win7MSIXInstaller/StartupTask.cpp
```diff
@@ -0,0 +1,193 @@
...
+        ZeroMemory(&variantNull, sizeof(VARIANT));
+        variantNull.vt = VT_NULL;
+
+        ComPtr<ITaskService> taskService;
+        RETURN_IF_FAILED(CoCreateInstance(CLSID_TaskScheduler,
+            NULL,
+            CLSCTX_INPROC_SERVER,
+            IID_ITaskService,
+            reinterpret_cast<void**>(&taskService)));
+        RETURN_IF_FAILED(taskService->Connect(variantNull, variantNull, variantNull, variantNull));
```
sarjanas-msft: >ariantNull, variantNull, variantNull, variantNull [](start = 47, length = 49)

comments to clarify all the Nulls

preview/Win7Msix/Win7MSIXInstaller/StartupTask.cpp
wcheng-msft: good catch, handling multiple of these

---
In reply to: [277072724](https://github.com/Microsoft/msix-packaging/pull/87#discussion_r277072724) [](ancestors = 277072724)

preview/Win7Msix/Win7MSIXInstaller/StartupTask.cpp
wcheng-msft: method summary comments are in the .hpp 

---
In reply to: [277070726](https://github.com/Microsoft/msix-packaging/pull/87#discussion_r277070726) [](ancestors = 277070726)

preview/Win7Msix/Win7MSIXInstaller/StartupTask.cpp
wcheng-msft: this is NOT compatible with win10 objects, as win10 has OS support for specifically allowing apps to enable/disable their startup tasks, and OS support to run the tasks as part of the logon sequence. This instead creates tasks using task scheduler because there is no other OS support in win7.

---
In reply to: [277071361](https://github.com/Microsoft/msix-packaging/pull/87#discussion_r277071361) [](ancestors = 277071361)

preview/Win7Msix/Win7MSIXInstaller/MsixRequest.cpp
wcheng-msft: i'll need to merge this in with Rudy's split change, but the reasoning here is that ErrorHandler (which does rollback) expects packageInfo to be filled in. early failures to open the package should not go to ErrorHandler.


---
In reply to: [277068659](https://github.com/Microsoft/msix-packaging/pull/87#discussion_r277068659) [](ancestors = 277068659)

preview/Win7Msix/Win7MSIXInstaller/StartupTask.cpp
wcheng-msft: updated the comment in .hpp to make this clear

---
In reply to: [277906358](https://github.com/Microsoft/msix-packaging/pull/87#discussion_r277906358) [](ancestors = 277906358,277071361)

PR #95, opened 2019-04-23
Lite Xmlwriter and start processing payload files

msftrubengu:
- Create lite XmlWriter class that doesn't use any SAX or DOM implementationd and just writes into a stream. This allows us to don't have to deal with a PAL for xml creation, memory concerns for big DOMs and compilation time used by xerces.
- Create BlockMapWriter and ContentTypeWriter classes that uses the new XmlWriter.
- Start processing payload files by getting its information and blocks.
- Fix a bug in DirectoryObject where the directory root was not correctly removed from the file name.

Next: Fully create the AppxBlockMap.xml file, which out the information required from the zip object (lfh)


src/inc/StringStream.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+        HRESULT STDMETHODCALLTYPE Write(const void *buffer, ULONG countBytes, ULONG *bytesWritten) noexcept override try
+        {
+            if (bytesWritten) { *bytesWritten = 0; }
+            ULONG oldOffset = m_offset;
+            m_data.write(static_cast<const char*>(buffer), static_cast<std::streamsize>(countBytes));
+            m_offset = std::min(m_offset + countBytes, static_cast<ULONG>(m_data.tellp()));
+            ThrowErrorIfNot(Error::FileWrite, ((m_offset - oldOffset) == countBytes), "write failed");
+            if (bytesWritten) { *bytesWritten = countBytes; }
+            m_data.seekp(m_offset);
```
JohnMcPMS: A lot of unneeded math. If min takes first, guaranteed to not throw.  If second, guaranteed to throw.

Equivalent would be:

write()
m_offset += countBytes;
ThrowIf(m_offset != m_data.tellp())
write to out param

JohnMcPMS: Even more unnecessary since I looked it up:  basic_ostream's already have a cursor that they track, with all the same functionality here.  There is no need to do anything but forward to it.

---
In reply to: [278220325](https://github.com/Microsoft/msix-packaging/pull/95#discussion_r278220325) [](ancestors = 278220325)

JohnMcPMS: The only tricky part will be avoiding the state bits that can get set, but that isn't really so bad.

---
In reply to: [278222223](https://github.com/Microsoft/msix-packaging/pull/95#discussion_r278222223) [](ancestors = 278222223,278220325)

src/inc/StringStream.hpp
```diff
@@ -0,0 +1,68 @@
...
+#pragma once
+
+#include <Exceptions.hpp>
+#include <StreamBase.hpp>
+
+#include <sstream>
+
+namespace MSIX {
+
+    class StringStream final : public StreamBase
```
JohnMcPMS: I would suggest a secondary interface or direct usage of this class to add a helper that simply takes a string for writing.  Or even a static/standalone helper to simplify the call sites.

src/msix/pack/XmlWriter.cpp
```diff
@@ -0,0 +1,83 @@
...
+        const static char* xmlStart = "<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"true\"?><";
+
+        // Adds xml header declaration plus the name of the root element
+        XmlWriter::XmlWriter(const std::string& root)
+        {
+            m_elements.emplace(root);
+            m_stream = ComPtr<IStream>::Make<StringStream>();
+            std::string start = xmlStart + root;
+            ULONG copy;
+            ThrowHrIfFailed(m_stream->Write(static_cast<const void*>(start.data()), static_cast<ULONG>(start.size()), &copy));
```
JohnMcPMS: >Write(static_cast<const void*>(start.data()), static_cast<ULONG>(start.size()), &copy) [](start = 38, length = 86)

RE: that standalone helper, this could just be:

WriteString(start)

src/msix/pack/XmlWriter.cpp
```diff
@@ -0,0 +1,83 @@
...
+            // tag and add the new one. If the state is closed, there is no need to close the tag as it had already been closed.
+            if (m_state == State::OpenElement)
+            {
+                // close parent element
+                element = "><" + name;
+            }
+            else // State::ClosedElement
+            {
+                element = "<" + name;
+            }
```
JohnMcPMS: nit: I find it easier to follow if any common code is not inside conditionals.  Thus:

if (open)
  >

< + name

src/msix/pack/XmlWriter.cpp
```diff
@@ -0,0 +1,83 @@
...
+            ULONG copy;
+            ThrowHrIfFailed(m_stream->Write(static_cast<const void*>(element.data()), static_cast<ULONG>(element.size()), &copy));
```
JohnMcPMS: With a simplified call to write a string, consider simply writing as you go rather than constructing a temporary.

src/msix/pack/XmlWriter.cpp
```diff
@@ -0,0 +1,83 @@
...
+            m_stream = ComPtr<IStream>::Make<StringStream>();
+            std::string start = xmlStart + root;
+            ULONG copy;
+            ThrowHrIfFailed(m_stream->Write(static_cast<const void*>(start.data()), static_cast<ULONG>(start.size()), &copy));
+            m_state = State::OpenElement;
+        }
+
+        void XmlWriter::StartElement(const std::string& name)
+        {
+            ThrowErrorIf(Error::XmlError, m_state == State::Finish, "Invalid call, xml already closed");
```
JohnMcPMS: >closed [](start = 95, length = 6)

finished

src/msix/pack/XmlWriter.cpp
```diff
@@ -0,0 +1,83 @@
...
+            m_elements.pop();
+            if (m_elements.size() == 0)
+            {
+                m_state = State::Finish;
+            }
+        }
+
+        void XmlWriter::AddAttribute(const std::string& name, const std::string& value)
+        {
+            ThrowErrorIf(Error::XmlError, (m_state == State::Finish) || (m_state == State::ClosedElement), "Invalid call to AddAttrbute");
```
JohnMcPMS: typo

JohnMcPMS: >AddAttrbute [](start = 124, length = 11)

typo

src/msix/pack/XmlWriter.cpp
```diff
@@ -0,0 +1,83 @@
...
+            // Name="Value". Always add space at the beginning.
+            std::string attribute = " " + name + "=\"" + value + "\"";
```
JohnMcPMS: >value [](start = 57, length = 5)

Should we consider doing any escaping here?  Or not needed for our narrow cases?

src/inc/ContentTypeWriter.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+    protected:
+        void AddDefault(const std::string& ext, const std::string& contentType);
+        void AddOverride(const std::string& file, const std::string& contentType);
+
+        // File extension to MIME value map that are added as default elements
+        // If the extension is already in the map and its content type is different,
+        // AddOverride is called.
+        std::map<std::string, std::string> m_defaultExtensions;
+        std::unique_ptr<XmlWriter> m_xmlWriter;
```
JohnMcPMS: Why a unique_ptr?

src/inc/ContentTypeWriter.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+#include <map>
+
+namespace MSIX {
+
+    class ContentTypeWriter final
+    {
+    public:
+        ContentTypeWriter();
+        ~ContentTypeWriter() {}
```
JohnMcPMS: Not needed, or if you must declare it, use "= default"

src/inc/XmlWriter.hpp
```diff
@@ -0,0 +1,44 @@
...
+        typedef enum
+        {
+            OpenElement = 1,
+            ClosedElement = 2,
+            Finish = 3
+        }
+        State;
+
+        XmlWriter(const std::string& root);
+        ~XmlWriter() {};
```
JohnMcPMS: >{}; [](start = 21, length = 3)

= default

src/inc/XmlWriter.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+        void StartElement(const std::string& name);
+        void CloseElement();
+        void AddAttribute(const std::string& name, const std::string& value);
+        State GetState() { return m_state; }
+        ComPtr<IStream> GetStream() { return m_stream; }
```
JohnMcPMS: Consider requiring Finish state to get stream.

src/inc/StringStream.hpp
JohnMcPMS: Actually, a standalone helper is best, so it could target any IStream.  Oh to have extension functions syntactic sugar.

---
In reply to: [278223918](https://github.com/Microsoft/msix-packaging/pull/95#discussion_r278223918) [](ancestors = 278223918)

src/inc/XmlWriter.hpp
```diff
@@ -0,0 +1,44 @@
...
+    public:
+        typedef enum
+        {
+            OpenElement = 1,
+            ClosedElement = 2,
+            Finish = 3
+        }
+        State;
+
+        XmlWriter(const std::string& root);
```
JohnMcPMS: In the future, we may want to be able to specify the target stream.

src/msix/pack/AppxBlockMapWriter.cpp
```diff
@@ -0,0 +1,73 @@
...
+        m_xmlWriter->AddAttribute(sizeAttribute, std::to_string(uncompressedSize));
+        m_xmlWriter->AddAttribute(nameAttribute, name);
+        m_xmlWriter->AddAttribute(lfhSizeAttribute, std::to_string(lfh));
+    }
+
+    // <Block Size="2948" Hash="ORIk+3QF9mSpuOq51oT3Xqn0Gy0vcGbnBRn5lBg5irM="/>
+    void BlockMapWriter::AddBlockElement(const std::string& hash, std::size_t size)
+    {
+        m_xmlWriter->StartElement(blockElement);
+        if(size != 0)
```
JohnMcPMS: >0 [](start = 19, length = 1)

I would suggest that this become the default block size, so that the detail is encapsulated.

src/msix/pack/AppxBlockMapWriter.cpp
```diff
@@ -0,0 +1,73 @@
...
+    {
+        // TODO: might need sway \ to //. For the blockmap we always use the windows separator.
+        m_xmlWriter->StartElement(fileElement);
+        m_xmlWriter->AddAttribute(sizeAttribute, std::to_string(uncompressedSize));
+        m_xmlWriter->AddAttribute(nameAttribute, name);
+        m_xmlWriter->AddAttribute(lfhSizeAttribute, std::to_string(lfh));
+    }
+
+    // <Block Size="2948" Hash="ORIk+3QF9mSpuOq51oT3Xqn0Gy0vcGbnBRn5lBg5irM="/>
+    void BlockMapWriter::AddBlockElement(const std::string& hash, std::size_t size)
```
JohnMcPMS: I would think an AddBlockElement that took the hash bytes themselves would be better.  Again, trying to encapsulate the encoding of the data within the class rather than forcing the caller to be aware.

src/inc/AppxBlockMapWriter.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+namespace MSIX {
+
+    class BlockMapWriter final
+    {
+    public:
+        BlockMapWriter();
+        ~BlockMapWriter() {}
+
+        void AddFileElement(const std::string& name, std::uint64_t uncompressedSize, std::uint32_t lfh);
```
JohnMcPMS: nit: Remove Element from these names.

src/inc/AppxPackageWriter.hpp
```diff
@@ -64,8 +83,14 @@ namespace MSIX {
         }
         WriterState;
 
+        std::unique_ptr<PayloadFile> BuildPayloadFile(const std::string& name, const ComPtr<IStream>& stream, 
+            const std::string& contentType, APPX_COMPRESSION_OPTION compressionOpt);
+        void ProcessPayloadFiles(const std::vector<std::unique_ptr<PayloadFile>>& files);
+
         WriterState m_state;
         ComPtr<IStream> m_outputStream;
+        std::unique_ptr<BlockMapWriter> m_blockMapWriter;
```
JohnMcPMS: Why so unique?

No need to make everything a unique_ptr.

src/msix/pack/AppxPackageWriter.cpp
```diff
@@ -64,7 +110,77 @@ namespace MSIX {
...
+            files.push_back(std::move(payloadFile));
+        }
+        ProcessPayloadFiles(files);
+
         NOTIMPLEMENTED
     } CATCH_RETURN();
-}
\ No newline at end of file
+
+    std::unique_ptr<PayloadFile> AppxPackageWriter::BuildPayloadFile(const std::string& name, const ComPtr<IStream>& stream,
```
JohnMcPMS: This won't scale; you are holding the entire uncompressed package in memory.

JohnMcPMS: Potentially, depending on the call pattern.  But even if not, you are holding entire uncompressed files in memory.

---
In reply to: [278237881](https://github.com/Microsoft/msix-packaging/pull/95#discussion_r278237881) [](ancestors = 278237881)

src/msix/pack/AppxPackageWriter.cpp
```diff
@@ -2,26 +2,60 @@
...
+        m_contentTypeWriter = std::make_unique<ContentTypeWriter>();
     }
 
     // IPackageWriter
     void AppxPackageWriter::Pack(const ComPtr<IDirectoryObject>& from)
     {
         ThrowErrorIf(Error::InvalidState, m_state != WriterState::Open, "Invalid package writer state");
         auto fileMap = from->GetFilesByLastModDate();
+        std::vector<std::unique_ptr<PayloadFile>> files;
+        for(const auto& file : fileMap)
```
JohnMcPMS: TODO: Special case footprint file types.

JohnMcPMS: Probably after other files so they don't mess up the default.  Unless that is why you put in the force override.

---
In reply to: [278241762](https://github.com/Microsoft/msix-packaging/pull/95#discussion_r278241762) [](ancestors = 278241762)

src/msix/pack/AppxPackageWriter.cpp
```diff
@@ -2,26 +2,60 @@
...
+        {
+            std::string ext = file.second.substr(file.second.find_last_of(".") + 1);
+            std::transform(ext.begin(), ext.end(), ext.begin(), ::tolower);
+            std::string contentType;
+            APPX_COMPRESSION_OPTION compressionOpt;
+            auto findExt = extToContentType.find(ext);
+            if (findExt == extToContentType.end())
+            {
+                // if it's not in the map these are the defaults
```
JohnMcPMS: The interface should be something more like:

const ContentTypeData& GetExtensionContentTypeData(string ext)

And it would handle the map lookup and the defaulting.

src/msix/pack/AppxPackageWriter.cpp
```diff
@@ -64,7 +110,77 @@ namespace MSIX {
...
+            return hash;
+        };
+        for(auto& file : files)
+        {
+            for(auto& block : file->fileBlocks)
+            {
+                auto bind = std::bind(computeHash, block->block);
+                block->hashValue = std::async(std::launch::async, bind);
+            }
+        }
```
JohnMcPMS: This isn't what we need to parallelize.  And we shouldn't worry about parallel implementation just yet.  Try to design to allow it, but don't need to do it.

src/msix/pack/XmlWriter.cpp
msftrubengu: I don't believe is necessary for our scenario.

---
In reply to: [278228018](https://github.com/Microsoft/msix-packaging/pull/95#discussion_r278228018) [](ancestors = 278228018)

src/msix/pack/AppxPackageWriter.cpp
msftrubengu: Actually GetFilesByLastModDate shouldn't return any footprint file. The only footprint file we look in a given directory to pack is the AppxManifest. Any other should be just ignore them. I'll rename it to GetPayloadFilesByLastModDate, we could us the StorageObject FileNameOptions enum, but I don't see any other use case that require this map with payload files

And yes, the force override is going to be used for the footprint files, specially with the AppxSignature.p7x

---
In reply to: [278242004](https://github.com/Microsoft/msix-packaging/pull/95#discussion_r278242004) [](ancestors = 278242004,278241762)

src/inc/XmlWriter.hpp
msftrubengu: Would it take ownership of the stream? I made a constructor that takes a ComPtr<IStream>&, if we decide that it shouldn't take ownership of it then I'll change it to IStream* and add ref it

---
In reply to: [278233393](https://github.com/Microsoft/msix-packaging/pull/95#discussion_r278233393) [](ancestors = 278233393)

src/inc/StringStream.hpp
msftrubengu: Helper in StreamHelper.hpp

---
In reply to: [278233164](https://github.com/Microsoft/msix-packaging/pull/95#discussion_r278233164) [](ancestors = 278233164,278223918)

src/inc/AppxBlockMapWriter.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+namespace MSIX {
+
+    class BlockMapWriter final
+    {
+    public:
+        BlockMapWriter();
+
+        void AddFile(const std::string& name, std::uint64_t uncompressedSize, std::uint32_t lfh);
+        void AddBlock(const std::vector<std::uint8_t>& hash, std::size_t size = 0);
```
JohnMcPMS: >std::size_t size = 0 [](start = 61, length = 20)

Shouldn't need size now; comes with vector.

src/inc/ContentType.hpp
```diff
@@ -0,0 +1,22 @@
...
+#include <map>
+
+namespace MSIX {
+
+    typedef std::pair<std::string, APPX_COMPRESSION_OPTION> ContentTypeData;
+
+    class ContentType
+    {
+    public:
+        static const ContentTypeData GetExtensionContentTypeData(std::string& ext);
```
JohnMcPMS: >ContentTypeData [](start = 21, length = 15)

Just my own stylistic here: if you are going to have these be static, they should return a ContentType.

src/inc/DirectoryObject.hpp
```diff
@@ -56,9 +56,18 @@ namespace MSIX {
 
         // IDirectoryObject
         ComPtr<IStream> OpenFile(const std::string& fileName, MSIX::FileStream::Mode mode) override;
-        std::multimap<std::uint64_t, std::string> GetFilesByLastModDate() override;
+        std::multimap<std::uint64_t, std::string> GetPayloadFilesByLastModDate() override;
 
     protected:
+        bool IsFootPrintFile(std::string normalized)
```
JohnMcPMS: >IsFootPrintFile [](start = 13, length = 15)

Not sure why this is here, but it doesn't seem like the responsibility of the directory object to filter package specific files.

src/inc/StreamHelper.hpp
```diff
@@ -49,5 +49,12 @@ namespace MSIX {
             ThrowHrIfFailed(stream->Seek(start, StreamBase::Reference::START, nullptr));
             return std::make_pair(streamSize, std::move(buffer));
         }
+
+        inline ULONG WriteStringToStream(const ComPtr<IStream>& stream, const std::string& toWrite)
+        {
+            ULONG written;
+            ThrowHrIfFailed(stream->Write(static_cast<const void*>(toWrite.data()), static_cast<ULONG>(toWrite.size()), &written));
+            return written;
```
JohnMcPMS: >written [](start = 19, length = 7)

I'd also consider not returning this and ThrowIf(not enough bytes written)

src/inc/StringStream.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+        HRESULT STDMETHODCALLTYPE Write(const void *buffer, ULONG countBytes, ULONG *bytesWritten) noexcept override try
+        {
+            if (bytesWritten) { *bytesWritten = 0; }
+            m_data.write(static_cast<const char*>(buffer), static_cast<std::streamsize>(countBytes));
+            // std::basic_ostream::write : Characters are inserted into the output sequence until one of the following occurs:
+            // exactly count characters are inserted or inserting into the output sequence fails (in which case setstate(badbit) is called)
+            // If the state is std::ios_base::goodbit we know the exact number of bytes were written.
+            ThrowErrorIf(Error::FileWrite, m_data.rdstate() != std::ios_base::goodbit, "Write failed");
+            m_offset += countBytes;
```
JohnMcPMS: You can expand this to Read and Seek to avoid having an m_offset member at all.

src/inc/XmlWriter.hpp
JohnMcPMS: Was more of a thought that a required change, but no, it shouldn't take ownership of the stream.

---
In reply to: [278361740](https://github.com/Microsoft/msix-packaging/pull/95#discussion_r278361740) [](ancestors = 278361740,278233393)

src/msix/pack/AppxBlockMapWriter.cpp
JohnMcPMS: Reactivating since no change or comment.

---
In reply to: [278234343](https://github.com/Microsoft/msix-packaging/pull/95#discussion_r278234343) [](ancestors = 278234343)

src/inc/ContentType.hpp
JohnMcPMS: Also, why not const ContentType&

---
In reply to: [278644277](https://github.com/Microsoft/msix-packaging/pull/95#discussion_r278644277) [](ancestors = 278644277)

src/msix/pack/ContentType.cpp
```diff
@@ -0,0 +1,137 @@
...
+            { "mpg",   { "video/mpeg", APPX_COMPRESSION_OPTION_NONE} },
+            { "mov",   { "video/quicktime", APPX_COMPRESSION_OPTION_NONE} },
+            { "wmv",   { "video/x-ms-wmv", APPX_COMPRESSION_OPTION_NONE} }
+        };
+
+        auto findExt = extToContentType.find(ext);
+        if (findExt == extToContentType.end())
+        {
+            // if it's not in the map these are the defaults
+            return std::make_pair("application/octet-stream", APPX_COMPRESSION_OPTION_NORMAL);
```
JohnMcPMS: >std::make_pair("application/octet-stream", APPX_COMPRESSION_OPTION_NORMAL); [](start = 19, length = 75)

Can return const& if this is also a static.

src/msix/pack/ContentType.cpp
```diff
@@ -0,0 +1,137 @@
...
+        if (footprintFile == APPX_FOOTPRINT_FILE_TYPE_BLOCKMAP)
+        {
+            return "application/vnd.ms-appx.blockmap+xml";
+        }
+        if (footprintFile == APPX_FOOTPRINT_FILE_TYPE_SIGNATURE)
+        {
+            return "application/vnd.ms-appx.signature";
+        }
+        // TODO: add other ones if needed, otherwise throw
+        NOTIMPLEMENTED
```
JohnMcPMS: >NOTIMPLEMENTED [](start = 8, length = 14)

NOTSUPPORTED

JohnMcPMS: Although a more verbose exception would probably be better.

---
In reply to: [278648510](https://github.com/Microsoft/msix-packaging/pull/95#discussion_r278648510) [](ancestors = 278648510)

src/msix/pack/ContentTypeWriter.cpp
```diff
@@ -0,0 +1,90 @@
...
+    static const char* typesElement = "Types";
+    static const char* typesNamespace = "http://schemas.openxmlformats.org/package/2006/content-types";
+    static const char* defaultElement = "Default";
+    static const char* contentTypeAttribute = "ContentType";
+    static const char* extensionAttribute = "Extension";
+    static const char* overrideElement = "Override";
+    static const char* partNameAttribute = "PartName";
+
+    // <Types xmlns="http://schemas.openxmlformats.org/package/2006/content-types">
+    ContentTypeWriter::ContentTypeWriter() : m_xmlWriter(XmlWriter(typesElement))
```
JohnMcPMS: >XmlWriter [](start = 57, length = 9)

Not needed.

src/msix/pack/ContentType.cpp
msftrubengu: Change to throw not support with a log message

---
In reply to: [278648651](https://github.com/Microsoft/msix-packaging/pull/95#discussion_r278648651) [](ancestors = 278648651,278648510)

src/msix/pack/ContentTypeWriter.cpp
msftrubengu: I don't see why this isn't needed. XmlWriter doesn't have a default constructor which is intentional

---
In reply to: [278648822](https://github.com/Microsoft/msix-packaging/pull/95#discussion_r278648822) [](ancestors = 278648822)

src/inc/DirectoryObject.hpp
msftrubengu: moved to packagewriter

---
In reply to: [278644877](https://github.com/Microsoft/msix-packaging/pull/95#discussion_r278644877) [](ancestors = 278644877)

PR #98, opened 2019-04-25
Add VS 2019 direct CMakeCache config

JohnMcPMS:
Adds the direct CMakeCache opening config, so that VS 2019 can just consume the output of makewin.cmd


JohnMcPMS:
Overriding policy as this is a simple config update.

PR #101, opened 2019-04-26
Display percentage on UI

jyvenugo:
This PR updates the UI to display, 'Installing app package (percent)%...' when app installation is in progress with the percentage value from the msixresponse object.
Updated spelling of g_height.


preview/Win7Msix/Win7MSIXInstaller/InstallUI.cpp
```diff
@@ -137,13 +137,20 @@ LRESULT CALLBACK WndProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
...
+                ::DeleteObject(hbr);
+                SetBkMode((HDC)wParam, TRANSPARENT);
+                return (LRESULT)::GetStockObject(NULL_BRUSH);
+            }
+            case IDC_STATICPERCENTCONTROL:
+            {
+                HBRUSH hbr = (HBRUSH)DefWindowProc(hWnd, message, wParam, lParam);
+                ::DeleteObject(hbr);
+                SetBkMode((HDC)wParam, TRANSPARENT);
+                return (LRESULT)::GetStockObject(NULL_BRUSH);
```
wcheng-msft: is this expected to differ later from IDC_LAUNCHCHECKBOX or would they always be the same?

jyvenugo: It is just the background color and formatting. Should be the same. Would it be better to remove the case and keep it common for all?

wcheng-msft: if there's some controls that should, and others that shouldn't then you could just have the first case __fallthrough; if all controls should, then you don't need the switch. 

---
In reply to: [279122987](https://github.com/Microsoft/msix-packaging/pull/101#discussion_r279122987) [](ancestors = 279122987)

jyvenugo: Removed switch case. Ran the installer and verified changes.

PR #102, opened 2019-04-26
Initial structure of the semantic validation

JohnMcPMS:
This change introduces the structure of the semantic validation code, which ensures certain internal consistencies that are harder to verify through a schema.  It also moves around a few things related to validation and XML to provide cleaner call sites.

Full semantic validation will only be enabled when the validating parser is, and thus the schemas have been checked first.  The existing checks inside AppxManifestObject et al will also stay as is for now.

Verified by:
makewin.cmd x64 -vp
test\win32\win32.ps1


src/msix/PAL/XML/xerces-c/XmlObject.cpp
```diff
@@ -251,13 +251,13 @@ class XercesElement final : public ComClass<XercesElement, IXmlElement, IXercesE
...
     {
-        auto attributeName = wstring_to_utf8(attributeNames[static_cast<uint8_t>(attribute)]);
+        auto attributeName = GetAttributeNameStringUtf8(attribute);
         return GetAttributeValue(attributeName);
     }
 
     std::vector<std::uint8_t> GetBase64DecodedAttributeValue(XmlAttributeName attribute) override
     {
-        XercesXMLChPtr nameAttr(XMLString::transcode(wstring_to_utf8(attributeNames[static_cast<uint8_t>(attribute)]).c_str()));
+        XercesXMLChPtr nameAttr(XMLString::transcode(GetAttributeNameStringUtf8(attribute).c_str()));
```
msftrubengu: >.c_str() [](start = 90, length = 8)

GetAttributeNameStringUtf8 returns const char*

PR #109, opened 2019-04-30
Validate multiple target device families in app manifest

jyvenugo:
There could be multiple target device families listed in an app manifest. Currently, the code does not check for compatibility of multiple TDF's. ValidateTargetDeviceFamily.cpp should return S_OK if any one of them is compatible with the OS.
If none of them are compatible, ERROR_INSTALL_REJECTED is returned.

### DIFF
--- a/preview/MsixCore/MsixCoreInstaller/ValidateTargetDeviceFamily.cpp
+++ b/preview/MsixCore/MsixCoreInstaller/ValidateTargetDeviceFamily.cpp
@@ -78,8 +78,18 @@ HRESULT ValidateTargetDeviceFamily::ParseAndValidateTargetDeviceFamilyFromPackag
         {
             TraceLoggingWrite(g_MsixTraceLoggingProvider,
                 "Target device family name and manifest min version are compatible with OS",
-                TraceLoggingLevel(WINEVENT_LEVEL_INFO));
+                TraceLoggingLevel(WINEVENT_LEVEL_INFO),
+                TraceLoggingValue(m_targetDeviceFamilyName.c_str(), "TargetDeviceFamilyName"),
+                TraceLoggingValue(manifestMinVersion.c_str(), "ManifestMinVersion"));
             return S_OK;
+        }
+        else
+        {
+            TraceLoggingWrite(g_MsixTraceLoggingProvider,
+                "Target device family name and manifest min version are not compatible with OS",
+                TraceLoggingLevel(WINEVENT_LEVEL_INFO),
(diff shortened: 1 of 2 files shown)


preview/MsixCore/MsixCoreInstaller/ValidateTargetDeviceFamily.cpp
```diff
@@ -58,33 +45,50 @@ HRESULT ValidateTargetDeviceFamily::ParseTargetDeviceFamilyFromPackage()
...
+        m_buildNumber = manifestMinVersion.substr(start, end - start);
+
+        /// Return if any one of the target device families are compatible with OS
+        if (IsTargetDeviceFamilyNameCompatibleWithOS() && IsManifestVersionCompatibleWithOS())
+        {
+            TraceLoggingWrite(g_MsixTraceLoggingProvider,
+                "Target device family name and manifest min version are compatible with OS",
+                TraceLoggingLevel(WINEVENT_LEVEL_INFO));
+            return S_OK;
+        }
```
wcheng-msft: >} [](start = 8, length = 1)

I think it would be helpful to add a tracelogging to tell us what values we found in the manifest that were not compatible.

jyvenugo: Added logging for values in either case.

preview/MsixCore/MsixCoreInstaller/ValidateTargetDeviceFamily.cpp
```diff
@@ -58,33 +45,60 @@ HRESULT ValidateTargetDeviceFamily::ParseTargetDeviceFamilyFromPackage()
...
+                TraceLoggingValue(manifestMinVersion.c_str(), "ManifestMinVersion"));
+            return S_OK;
+        }
+        else
+        {
+            TraceLoggingWrite(g_MsixTraceLoggingProvider,
+                "Target device family name and manifest min version are not compatible with OS",
+                TraceLoggingLevel(WINEVENT_LEVEL_INFO),
+                TraceLoggingValue(m_targetDeviceFamilyName.c_str(), "TargetDeviceFamilyName"),
+                TraceLoggingValue(manifestMinVersion.c_str(), "ManifestMinVersion"));
```
wcheng-msft: >manifestMinVersion [](start = 34, length = 18)

manifestMinVersion gets truncated as you process it, the original string should be minVersion.Get()

jyvenugo: Updated.

PR #122, opened 2019-05-07
[MsixCore] Support installation of packages via IStream

### LINKED ISSUE #121: [FEATURE REQUEST] Msix Core should allow the installation of Msix packages passed as stream.
*Project*
Msix Core

### Describe the feature
In addition to the 2 existing functions `AddPackage(wstring filepath...)` and `AddPackageAsync(wstring filepath....)` using a file path to represent a msix package file, MsixCore should also support the installation of Msix packages not stored as files but as streams.

### Usage

Some examples:
- Install a package downloaded from internet (Http stream)
- Install a package embedded in a executable as a resource (self-extrator installer)
- Install a package from a zip stream
...

rudyhuyn:
Fix #121

> In addition to the 2 existing functions AddPackage(wstring filepath...) and AddPackageAsync(wstring filepath....) using a file path to represent a msix package file, MsixCore should also support the installation of Msix packages not stored as files but as streams.

### Changes

Modify `AddPackage` and `AddPackageAsync` to replace the file path by a `IStream`:

```cpp
  HRESULT AddPackage(IStream * packageStream, DeploymentOptions options);`
```

```cpp
  std::shared_ptr<IMsixResponse> AddPackageAsync(IStream * packageStream, DeploymentOptions options, std::function<void(const IMsixResponse&)> callback = nullptr);`
```

And add 2 methods with the same prototype than the 2 original methods to still support installation of packages based on a file path.

MsixRequest was also modified to use a `IStream` instead of a file path as a reference of the Msix package we want to install.

### DIFF
--- a/preview/MsixCore/MsixCoreInstallerLib/inc/IPackageManager.hpp
+++ b/preview/MsixCore/MsixCoreInstallerLib/inc/IPackageManager.hpp
@@ -6,13 +6,14 @@
 #include "IPackage.hpp"
 #include "DeploymentOptions.hpp"
 
-
 namespace MsixCoreLib {
     class IPackageManager
     {
     public:
         virtual std::shared_ptr<IMsixResponse> AddPackageAsync(const std::wstring & packageFilePath, DeploymentOptions options, std::function<void(const IMsixResponse&)> callback = nullptr) = 0;
+        virtual std::shared_ptr<IMsixResponse> AddPackageAsync(IStream * packageStream, DeploymentOptions options, std::function<void(const IMsixResponse&)> callback = nullptr) = 0;
         virtual HRESULT AddPackage(const std::wstring & packageFilePath, DeploymentOptions options) = 0;
+        virtual HRESULT AddPackage(IStream * packageStream, DeploymentOptions options) = 0;
         virtual std::shared_ptr<IMsixResponse> RemovePackageAsync(const std::wstring & packageFullName, std::function<void(const IMsixResponse&)> callback = nullptr) = 0;
         virtual HRESULT RemovePackage(const std::wstring & packageFullName) = 0;
         virtual std::shared_ptr<IInstalledPackage> FindPackage(const std::wstring & packageFamilyName) = 0;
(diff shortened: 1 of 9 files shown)


preview/MsixCore/MsixCoreInstallerLib/inc/IPackageManager.hpp
```diff
@@ -6,13 +6,14 @@
...
 
-
 namespace MsixCoreLib {
     class IPackageManager
     {
     public:
         virtual std::shared_ptr<IMsixResponse> AddPackageAsync(const std::wstring & packageFilePath, DeploymentOptions options, std::function<void(const IMsixResponse&)> callback = nullptr) = 0;
+        virtual std::shared_ptr<IMsixResponse> AddPackageAsync(IStream * packageStream, DeploymentOptions options, std::function<void(const IMsixResponse&)> callback = nullptr) = 0;
         virtual HRESULT AddPackage(const std::wstring & packageFilePath, DeploymentOptions options) = 0;
+        virtual HRESULT AddPackage(IStream * packageStream, DeploymentOptions options) = 0;
```
wcheng-msft: >IStream [](start = 35, length = 7)

can you give us some context or insight on why you wanted to support IStream on this interface as well. i.e. what scenarios are you thinking about where filepath by itself insufficient?

while I think the changes are fine, one concern is that
win10 PackageManager doesn't expose any interfaces that take an IStream. We want installs of MSIX packages to go through the appxdeploymentserver stack on OS's that support MSIX (RS3+) so taking in a stream here means we'd have to download the package to a file in order to pass onto the RS3+ PackageManager APIs.

rudyhuyn: My main goal is to be able to create a self-extractor installer, with a msix package embedded as a resource, so instead of extracting the resource, save it to %Temp% and call AddPackage, we can directly install the msix from the resource stream.

"We want installs of MSIX packages to go through the appxdeploymentserver stack on OS's that support MSIX (RS3+)"

This is interesting to know! It really depends where you want to add the abstraction layer and how you plan to manage this integration. In MsixInstaller.exe or in MsixInstallerLib.lib? I suppose in the executable, if it's the case, it should not be an issue, otherwise, I agree, it would be problematic.

wcheng-msft: We discussed internally, yes we do want both the exe and the lib to be able to redirect to PackageManager. So it seems like for the embedded resource case we'd have to extract it to a temp file and call AddPackage from within the lib.

PR #135, opened 2019-05-15
Minor pack improvements

msftrubengu:
- Add MSIX_PACKUNPACK_OPTION flat to PackPackage entrypoint. This allows us to follow the same patter as UnpackPackage and be open future changes are not validation switches instead of creating a new entrypoint. For example, we might allow naming the package after the package id from the manifest. 
- Move SHA256 hashing inside the BlockMapWriter. I realized that there was need to the AppxPackageWriter to the hashing, just pass the block to the block map writer and encapsulate the SHA256 hashing and the base64 encoding.
- Add compiler and linker flags to hide visibility and discard unused functions for Android. This result in a 10KB drop in the binary size.

### DIFF
--- a/src/msix/pack/AppxBlockMapWriter.cpp
+++ b/src/msix/pack/AppxBlockMapWriter.cpp
@@ -57,9 +57,9 @@ namespace MSIX {
     {
         // hash block
         std::vector<std::uint8_t> hash;
-        ThrowErrorIfNot(MSIX::Error::SignatureInvalid, 
+        ThrowErrorIfNot(MSIX::Error::BlockMapInvalidData,
             MSIX::SHA256::ComputeHash(const_cast<std::uint8_t*>(block.data()), static_cast<uint32_t>(block.size()), hash), 
-            "Invalid signature");
+            "Failed computing hash");
 
         m_xmlWriter.StartElement(blockElement);
         m_xmlWriter.AddAttribute(hashAttribute, Base64::ComputeBase64(hash));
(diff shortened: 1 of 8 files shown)


src/msix/pack/AppxBlockMapWriter.cpp
```diff
@@ -53,8 +53,14 @@ namespace MSIX {
     }
 
     // <Block Size="2948" Hash="ORIk+3QF9mSpuOq51oT3Xqn0Gy0vcGbnBRn5lBg5irM="/>
-    void BlockMapWriter::AddBlock(const std::vector<std::uint8_t>& hash, ULONG size, bool isCompressed)
+    void BlockMapWriter::AddBlock(const std::vector<std::uint8_t>& block, ULONG size, bool isCompressed)
     {
+        // hash block
+        std::vector<std::uint8_t> hash;
+        ThrowErrorIfNot(MSIX::Error::SignatureInvalid,
```
JohnMcPMS: >SignatureInvalid [](start = 37, length = 16)

Not the correct error to use; might need a new one.  Or maybe a generic "invalid data" error here.

src/msix/pack/AppxBlockMapWriter.cpp
```diff
@@ -53,8 +53,14 @@ namespace MSIX {
...
 
     // <Block Size="2948" Hash="ORIk+3QF9mSpuOq51oT3Xqn0Gy0vcGbnBRn5lBg5irM="/>
-    void BlockMapWriter::AddBlock(const std::vector<std::uint8_t>& hash, ULONG size, bool isCompressed)
+    void BlockMapWriter::AddBlock(const std::vector<std::uint8_t>& block, ULONG size, bool isCompressed)
     {
+        // hash block
+        std::vector<std::uint8_t> hash;
+        ThrowErrorIfNot(MSIX::Error::SignatureInvalid, 
+            MSIX::SHA256::ComputeHash(const_cast<std::uint8_t*>(block.data()), static_cast<uint32_t>(block.size()), hash), 
+            "Invalid signature");
```
JohnMcPMS: >"Invalid signature" [](start = 12, length = 19)

Not the correct string

PR #136, opened 2019-05-17
FTA handler changes

jyvenugo:
Made changes to FTA, Com, ARP, Protocol handler such that the keys are deleted from the virtual mount if present before copying rest of the keys over to the actual registry.


preview/MsixCore/MsixCoreInstaller/RegistryDevirtualizer.hpp
```diff
@@ -23,17 +23,22 @@ class RegistryDevirtualizer
...
-    static HRESULT Create(_In_ std::wstring hiveFileName, _In_ MsixRequest* msixRequest, _Out_ RegistryDevirtualizer** instance);
+    static HRESULT Create(_In_ std::wstring hiveFileName, _In_ MsixRequest* msixRequest, _Out_ std::shared_ptr<RegistryDevirtualizer>* instance);
 
     /// Creates a GUID string as a temporary registry key's name
     /// The Registry.dat hive will be loaded under this name to avoid conflict with existing keys or other installs
     ///
     /// @param tempName - string containing the name
     static HRESULT CreateTempKeyName(std::wstring & tempName);
 
-    ~RegistryDevirtualizer();
```
wcheng-msft: >RegistryDevirtualizer [](start = 5, length = 21)

you need to have this unload the hive (if it wasn't already unloaded) so that in error cases the hive gets unloaded.

preview/MsixCore/MsixCoreInstaller/RegistryDevirtualizer.cpp
```diff
@@ -68,6 +68,54 @@ HRESULT RegistryDevirtualizer::Run(_In_ bool remove)
...
+    // Per user mapping
+    RegistryKey userClassesKey;
+    HRESULT hrOpenUserClassesKey = m_rootKey.OpenSubKey(virtualKeyPath, KEY_READ, &userClassesKey);
+    if (SUCCEEDED(hrOpenUserClassesKey))
+    {
+        RegistryKey extensionKey;
+        HRESULT hrUserExtensionKey = userClassesKey.OpenSubKey(extensionName.c_str(), KEY_READ, &extensionKey);
+        if (SUCCEEDED(hrUserExtensionKey))
+        {
+            HRESULT deleteUserKey = userClassesKey.DeleteSubKey(extensionName.c_str());
```
wcheng-msft: >deleteUserKey [](start = 20, length = 13)

probably want to log this if it fails

preview/MsixCore/MsixCoreInstaller/RegistryDevirtualizer.cpp
```diff
@@ -68,6 +68,54 @@ HRESULT RegistryDevirtualizer::Run(_In_ bool remove)
...
+        }
+    }
+
+    // Per user mapping
+    RegistryKey userClassesKey;
+    HRESULT hrOpenUserClassesKey = m_rootKey.OpenSubKey(virtualKeyPath, KEY_READ, &userClassesKey);
+    if (SUCCEEDED(hrOpenUserClassesKey))
+    {
+        RegistryKey extensionKey;
+        HRESULT hrUserExtensionKey = userClassesKey.OpenSubKey(extensionName.c_str(), KEY_READ, &extensionKey);
```
wcheng-msft: >KEY_READ [](start = 86, length = 8)

test this out, but i think you need KEY_WRITE in order to delete.

preview/MsixCore/MsixCoreInstaller/RegistryDevirtualizer.cpp
```diff
@@ -68,6 +68,54 @@ HRESULT RegistryDevirtualizer::Run(_In_ bool remove)
...
+
+    // map real root hive to virtual root hive
+    PCWSTR virtualKeyPath = L"";
+    for (auto mapping : mappings)
+    {
+        if (mapping.actualRootHive == rootHive && mapping.actualSubkey == subKeyPath)
+        {
+            virtualKeyPath = mapping.virtualKey;
+        }
+    }
```
wcheng-msft: this approach doesn't really work for when HKLM is the rootHive and the subKeyPath is anything besides L""

preview/MsixCore/MsixCoreInstaller/FileTypeAssociation.cpp
```diff
@@ -304,6 +282,13 @@ HRESULT FileTypeAssociation::ProcessFtaForAdd(Fta& fta)
         RETURN_IF_FAILED(verbCommandKey.SetStringValue(L"", verbCommand));
     }
 
+    for (auto extensionName = fta.extensions.begin(); extensionName != fta.extensions.end(); ++extensionName)
+    {
+        m_msixRequest->GetRegistryDevirtualizer()->DeleteKeyIfPresent(rootHive, classesKeyPath.c_str(), *extensionName);
+    }
+
+    m_msixRequest->GetRegistryDevirtualizer()->DeleteKeyIfPresent(rootHive, classesKeyPath.c_str(), fta.progID);
```
wcheng-msft: >progID [](start = 104, length = 6)

the progID that we create here may not be what's in the virtual hive. This needs to be read off of the virtual key. i.e. this is a author-specified arbitrary string.

for instance in VLC, .mp4 key says it's associated with VLC.mp4 (the progID) so they've created a VLC.mp4 key that specifies the information.
our installer will associate .mp4 with MsixCoreVLC.mp4, so we've created a MsixCoreVLC.mp4 key that specifies the information.

preview/MsixCore/MsixCoreInstaller/FileTypeAssociation.cpp
```diff
@@ -220,44 +220,22 @@ HRESULT ConvertLogoToIcon(std::wstring logoPath, std::wstring & iconPath)
     return S_OK;
 }
 
-HRESULT FileTypeAssociation::ProcessFtaForAdd(Fta& fta)
+HRESULT FileTypeAssociation::ProcessFtaForAdd(Fta& fta, HKEY rootHive)
```
wcheng-msft: >HKEY rootHive [](start = 56, length = 13)

this rootHive specifies where WE (the msixcore installer) are writing the keys. but regardless of where we write the keys, we want to delete from both HKCU/HKLM virtual locations

preview/MsixCore/MsixCoreInstaller/RegistryDevirtualizer.cpp
```diff
@@ -68,6 +68,54 @@ HRESULT RegistryDevirtualizer::Run(_In_ bool remove)
...
+        HRESULT hrUserExtensionKey = userClassesKey.OpenSubKey(extensionName.c_str(), KEY_READ, &extensionKey);
+        if (SUCCEEDED(hrUserExtensionKey))
+        {
+            HRESULT deleteUserKey = userClassesKey.DeleteSubKey(extensionName.c_str());
+        }
+    }
+
+    //Per machine key
+    RegistryKey machineClassesKey;
+    HRESULT hrOpenMachineClassesKey = m_rootKey.OpenSubKey(L"MACHINE\\Software\\Classes", KEY_READ, &machineClassesKey);
```
wcheng-msft: >Software\\Classes [](start = 70, length = 17)

this is hardcoded for Software\\Classes, so it works for FTA, but i'd want this to be more generic, so that it could work for uninstall key.

preview/MsixCore/MsixCoreInstaller/PrepareDevirtualizedRegistry.cpp
```diff
@@ -25,9 +25,11 @@ HRESULT PrepareDevirtualizedRegistry::ExtractRegistry(bool remove)
 {
     std::wstring registryFilePath = m_msixRequest->GetPackageDirectoryPath() + registryDatFile;
```
wcheng-msft: >registryFilePath [](start = 17, length = 16)

we may want to make a copy of this and store it somewhere so that we preserve the original copy.

preview/MsixCore/MsixCoreInstaller/FileTypeAssociation.cpp
```diff
@@ -335,7 +320,7 @@ HRESULT FileTypeAssociation::ProcessFtaForRemove(Fta& fta)
     for (auto extensionName = fta.extensions.begin(); extensionName != fta.extensions.end(); ++extensionName)
     {
         bool registryHasExtension = false;
-        RETURN_IF_FAILED(m_registryDevirtualizer->HasFTA(*extensionName, registryHasExtension));
+        RETURN_IF_FAILED(m_msixRequest->GetRegistryDevirtualizer()->HasFTA(*extensionName, registryHasExtension));
```
wcheng-msft: >RETURN_IF_FAILED(m_msixRequest->GetRegistryDevirtualizer()->HasFTA(*extensionName, registryHasExtension)); [](start = 8, length = 106)

because ProcessFtaForAdd always adds whatever's in the manifest, we shouldn't filter these out on uninstall.

jyvenugo: Removed call to HasFTA method, modified ProcessFTAForRemove so that all extensions are removed followed by progid.

preview/MsixCore/MsixCoreInstaller/RegistryDevirtualizer.cpp
jyvenugo: Modified to remove uninstall key passed from ARP handler as well. Appended subpath passed to end of the generated key so that it is more generic.

preview/MsixCore/MsixCoreInstaller/FileTypeAssociation.cpp
jyvenugo: Removed this parameter. Modified to check at both HKCU and HKLM locations in the virtual mount irrespective of where the key gets written.

preview/MsixCore/MsixCoreInstaller/FileTypeAssociation.cpp
jyvenugo: Modified to get the default key value(progId) for an extension in a FTA. Then passed this over to DeleteKeyIfPresent to delete it from either the user or machine locations in the virtual mount.

preview/MsixCore/MsixCoreInstaller/RegistryDevirtualizer.cpp
jyvenugo: Modified approach to generate virtual mount path based on subpath passed and not the mapping table.

preview/MsixCore/MsixCoreInstaller/RegistryDevirtualizer.cpp
jyvenugo: Updated to KEY_WRITE. Tested that key gets deleted with this in the debugger.

preview/MsixCore/MsixCoreInstaller/RegistryDevirtualizer.hpp
jyvenugo: Added destructor to unload the virtual hive in error cases.

wcheng-msft:
>     }

this can just call UnloadMountedHive. make sure the success case doesn't try to unload it twice (can set m_hiveFileNameExists = false after unloading it successfully)

---
Refers to: preview/MsixCore/MsixCoreInstaller/RegistryDevirtualizer.cpp:598 in e635302. [](commit_id = e635302e8ebb02584950a742545c4cfd39248e32, deletion_comment = False)

preview/MsixCore/MsixCoreInstaller/RegistryDevirtualizer.cpp
```diff
@@ -68,6 +68,114 @@ HRESULT RegistryDevirtualizer::Run(_In_ bool remove)
...
+        if (SUCCEEDED(hrExtensionKey))
+        {
+            HRESULT hrDeleteTree = virtualKey.DeleteTree(extensionName.c_str());
+            if (FAILED(hrDeleteTree))
+            {
+                /*TraceLoggingWrite(g_MsixTraceLoggingProvider,
+                    "Unable to delete extension",
+                    TraceLoggingLevel(WINEVENT_LEVEL_WARNING),
+                    TraceLoggingValue(hrDeleteTree, "HR"),
+                    TraceLoggingValue(extensionName->c_str(), "Extension"));*/
```
wcheng-msft: >-> [](start = 51, length = 2)

uncomment? -> should be .

preview/MsixCore/MsixCoreInstaller/PrepareDevirtualizedRegistry.cpp
```diff
@@ -25,9 +25,11 @@ HRESULT PrepareDevirtualizedRegistry::ExtractRegistry(bool remove)
 {
     std::wstring registryFilePath = m_msixRequest->GetPackageDirectoryPath() + registryDatFile;
 
-    AutoPtr<RegistryDevirtualizer> registryDevirtualizer;
+    std::shared_ptr<RegistryDevirtualizer> registryDevirtualizer;
     RETURN_IF_FAILED(RegistryDevirtualizer::Create(registryFilePath, m_msixRequest, &registryDevirtualizer));
-    RETURN_IF_FAILED(registryDevirtualizer->Run(remove));
+    m_msixRequest->SetRegistryDevirtualizer(registryDevirtualizer);
+
+    //RETURN_IF_FAILED(registryDevirtualizer->Run(remove));
```
wcheng-msft: >/RETURN_IF_FAILED(registryDevirtualizer->Run(remove)) [](start = 5, length = 53)

you can delete this.

preview/MsixCore/MsixCoreInstaller/RegistryDevirtualizer.cpp
jyvenugo: Updated to use '.'

preview/MsixCore/MsixCoreInstaller/PrepareDevirtualizedRegistry.cpp
jyvenugo: Moved this to ExecuteForRemoveRequest() in PrepareDevirtualizedRegistry.cpp since this should still run in the beginning for remove. Removed this line from ExtractRegistry since it should not run on add now.

preview/MsixCore/MsixCoreInstaller/RegistryDevirtualizer.cpp
jyvenugo: Logged.

preview/MsixCore/MsixCoreInstaller/PrepareDevirtualizedRegistry.cpp
jyvenugo: Copied registry.dat file to registryFileCopy.dat and passed copy file to other handlers.

jyvenugo:
Modified UnloadMountedHive method to set m_hiveFileNameExists to false in case the hive was successfully unloaded(ERROR_SUCCESS) so that it won't be unloaded a second time despite which ever case it gets called from for the first time.

PR #142, opened 2019-05-24
Update readme

jyvenugo:
Update Target Device Family to MsixCore.Desktop for Windows 7 and 8.

### DIFF
--- a/tdf-guidance.md
+++ b/tdf-guidance.md
@@ -107,12 +107,12 @@ The following table shows the target device families to declare in the manifest:
   <tr>
     <td class="tg-yw4l" rowspan="2">Windows</td>
     <td class="tg-yw4l">7</td>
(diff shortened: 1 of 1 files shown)


tdf-guidance.md
```diff
@@ -107,12 +107,12 @@ The following table shows the target device families to declare in the manifest:
   <tr>
     <td class="tg-yw4l" rowspan="2">Windows</td>
     <td class="tg-yw4l">7</td>
-    <td class="tg-baqh" colspan="2">Windows7.Desktop</td>
+    <td class="tg-baqh" colspan="2">MsixCore.Desktop</td>
```
msftrubengu: ><td class="tg-baqh" colspan="2">MsixCore.Desktop</td> [](start = 4, length = 53)

I think you can avoid having two td tags with MsixCore.Desktop by adding rowspan="2" here and deleting the other one

PR #149, opened 2019-05-30
msixtest - use catch2 as our test infrastructure for all supported platforms

msftrubengu:
Before this change, we had three different implementations to perform our tests:
1. Powershell script, for windows
2. Shell script for Linux and macOS
3. Common shared library used for Android and iOS apps

This change removes the overhead of adding a test in three times by creating a single implementation by using Catch2.

msixtest is either an executable or a shared library, depending on the platform. It has a single entrypoint `msixtest_main` that takes argc and argv, as main, plus the path were the test packages are located. The shared library is used for our mobile test apps, while non-mobile just forwards the arguments to `msixtest_main`.

There is no custom command added, so msixtest has the same commands as Catch2. See [here](https://github.com/catchorg/Catch2/blob/master/docs/command-line.md)

Some usage examples:
msixtest : Runs all the tests. By default, catch2 doesn't show results of all success tests. It just shows failures and the number of test cases that passed.
msixtest -s : Runs all the tests and show results of all success tests
msixtest <test name> : Runs only the test specified
msixtest [tag] : Runs all tests in an specific tag, eg. [unpack] runs all the unpack tests.

Test:
- There are currently 66 tests.
- 1 test is specific for "no bundle".
- 2 tests are disabled, Api_AppxManifestReader_Capabilities and Unpack_SignedTamperedBlockMap-TRUST_E_BAD_DIGEST_sv. See comments on api_package.cpp and unpack.cpp for the reasons. 

Pipelines:
- Before this change, when the test task failed, we had to look into the logs and scroll until the tests are added. With these new tests we can use Catch2 reporters + Azure DevOps Test to provide a much better experience. All of our pipelines had been updated to pass `-r junit -o TEST-MsixSDK-$(_artifact).xml` to msixtest in the test task and a new publish test task to publish the result xml. I used junit because is a format that both Azure and Catch2 understand. 
- "No bundle" builds had been added to all platforms and are fully tested now. Before this change we didn't had an automated way to test "no bundle support"

Other related changes:
- Add Catch v2.7.2 license to the THIRD PARTY CODE NOTICE
- Move /test to /src/test to make it specific of the MSIX SDK codebase and to not be confused with MsixCore.
- Move samples binaries to their own output folder for clarity.
- Removes redundant BINARY_NAME CMake variable in CMakeLists.txt
- Make utf8_to_wstring and wstring_to_utf8 more strict for Windows.
- Fix bug were a file path with international characters couldn't be read on windows when CreateStreamOnFile is used


src/inc/FileStream.hpp
```diff
@@ -23,12 +23,10 @@ namespace MSIX {
             static const char* modes[] = { "rb", "wb", "ab", "r+b", "w+b", "a+b" };
             #ifdef WIN32
             errno_t err = fopen_s(&m_file, name.c_str(), modes[mode]);
-            std::ostringstream builder;
-            builder << "file: '" << name << "' does not exist.";
-            ThrowErrorIfNot(Error::FileOpen, (err==0), builder.str().c_str());
+            ThrowErrorIfNot(Error::FileOpen, (err==0), std::string("file: " + m_name + "does not exists.").c_str());
```
JohnMcPMS: >s [](start = 102, length = 1)

nit: exist

src/msix/UnicodeConversion.cpp
```diff
@@ -22,10 +23,12 @@ namespace MSIX {
 
     std::wstring utf8_to_wstring(const std::string& utf8string)
     {
+        if (utf8string.empty()) { return {}; }
         #ifdef WIN32
-        int size = MultiByteToWideChar(CP_UTF8, 0, utf8string.data(), utf8string.size(), nullptr, 0);
+        int size = MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS, utf8string.data(), static_cast<int>(utf8string.size()), nullptr, 0);
+        ThrowLastErrorIf(size == 0, "Error converting to wstring");
```
JohnMcPMS: >size [](start = 25, length = 4)

There is an issue here that can be fixed in another change.  This size value includes the null char and thus the returned string does too.

msftrubengu: MSDN is confusing... first call has cbMultiByte as a positive integer which means: "the function processes exactly the specified number of bytes. If the provided size does not include a terminating null character, the resulting Unicode string is not null-terminated, and the returned length does not include this character". So it won't include the null char.

---
In reply to: [289088558](https://github.com/Microsoft/msix-packaging/pull/149#discussion_r289088558) [](ancestors = 289088558)

PR #150, opened 2019-05-31
Added tests for -pfn-flat and -extract-all options. 

stephenk-msft:
Added tests for -pfn-flat and -extract-all options:

For bundles:
1. -extract-all
2. -pfn -extract-all
3. -pfn-flat
4. -pfn-flat -extract-all

For packages:
1. -pfn-flat

Renamed -sa option as -extract-all


src/test/msixtest/unbundle.cpp
```diff
@@ -105,11 +105,122 @@ TEST_CASE("Unbundle_StoreSigned_Desktop_x86_x64_MoviesTV_lang_applicability_off"
...
+
+    // Verify all the files extracted on disk are correct
+    auto files = MsixTest::Unbundle::GetExpectedBundleFilesPfnFlatFullApplicable();
+    CHECK(MsixTest::Directory::CompareDirectory(outputDir, files));
+
+    std::cout << "Expected Files Present" << '\n';
+
+    // Verify non expected files aren't present
+    auto filesNotApplicable = MsixTest::Unbundle::GetExpectedFilesNoApplicable();
+    CHECK(!MsixTest::Directory::CompareDirectory(outputDir, filesNotApplicable));
```
JohnMcPMS: >!MsixTest::Directory::CompareDirectory [](start = 10, length = 38)

This is a dubious check; but I know it is used other places in this file.  In reality, we would want CompareDirectory to return an enum that indicates {no files, some files, all files} and then we can make better judgements here.

src/test/msixtest/unbundle.cpp
```diff
@@ -56,7 +56,7 @@ TEST_CASE("Unbundle_StoreSigned_Desktop_x86_x64_MoviesTV", "[unbundle]")
     CHECK(MsixTest::Directory::CompareDirectory(outputDir, files));
 
     // Verify non expected files aren't present
-    auto filesNotApplicable = MsixTest::Unbundle::GetExpectedFilesNoApplicable();
+    auto filesNotApplicable = MsixTest::Unbundle::GetExpectedFilesNoLanguageApplicable();
```
msftrubengu: >GetExpectedFilesNoLanguageApplicable [](start = 50, length = 36)

Why is this one NoLanguageApplicable? Shouldn't unbundle with full applicability also avoid unpacking scale?

stephenk-msft: You're right, thanks for catching this!

src/test/msixtest/unbundle.cpp
msftrubengu: Yeah actually this entire check is completely busted and is  totally my fault. This is always going succeed if the fist file in map doesn't exists in the directory. 
If you want to fix it, you can go with what John suggested? Let me know if you need help with it 

---
In reply to: [289545863](https://github.com/Microsoft/msix-packaging/pull/150#discussion_r289545863) [](ancestors = 289545863)

JohnMcPMS: Actually, just get rid of the entire check, as the first call verifies that the exact map of files is present, so verifying that some other set of files isn't present is not needed.  Can also delete the other instance of using !CompareDirectory.

PR #151, opened 2019-06-03
Display error on UI in case of return error

jyvenugo:


preview/MsixCore/MsixCoreInstaller/InstallUI.cpp
```diff
@@ -660,6 +693,23 @@ void UI::UpdateDisplayPercent(float displayPercent)
...
+    ShowWindow(g_staticPercentText, SW_HIDE);
+    ShowWindow(g_progressHWnd, SW_HIDE);
+    ShowWindow(g_checkboxHWnd, SW_HIDE);
+    ShowWindow(g_CancelbuttonHWnd, SW_HIDE);
+    
+    //Show Error Box
+    ShowWindow(g_staticErrorTextHWnd, SW_SHOW);
+    ShowWindow(g_staticErrorDescHWnd, SW_SHOW);
+    std::wstringstream errorDescription;
+    errorDescription << "App installation failed with error message: An internal error occurred with error " << hr << ".";
```
wcheng-msft: ><< [](start = 109, length = 2)

use std::hex to show this in hex instead of negative number

wcheng-msft: >"App installation failed with error message: An internal error occurred with error  [](start = 24, length = 83)

i think best to make this a string resource?

jyvenugo: Modified to use string resource.

jyvenugo: Modified to use std::hex.

PR #152, opened 2019-06-07
Added APIs to unpack from package reader

stephenk-msft:
Added new APIs to msix.dll that allow you to unpack a package or bundle by passing in a package/bundle reader. The purpose of this change is to enable the tooling component of the WVD app attach feature. By having the package reader created by the caller instead of internally by the UnpackPackage API, the caller can access the package manifests, determine the package full names of packages, and figure out the names of package folders created during the unpack process.


src/msix/msix.cpp
```diff
@@ -70,6 +70,26 @@ MSIX_API HRESULT STDMETHODCALLTYPE UnpackPackage(
     return static_cast<HRESULT>(MSIX::Error::OK);
 } CATCH_RETURN();
 
+MSIX_API HRESULT STDMETHODCALLTYPE UnpackPackageFromPackageReader(
+    IAppxPackageReader* packageReader,
+    MSIX_PACKUNPACK_OPTION packUnpackOptions,
+    MSIX_VALIDATION_OPTION validationOption,
```
msftrubengu: >MSIX_VALIDATION_OPTION validationOption, [](start = 4, length = 40)

The validation options are just for the factory, you don't need them.

src/msix/msix.cpp
```diff
@@ -70,6 +70,26 @@ MSIX_API HRESULT STDMETHODCALLTYPE UnpackPackage(
...
 } CATCH_RETURN();
 
+MSIX_API HRESULT STDMETHODCALLTYPE UnpackPackageFromPackageReader(
+    IAppxPackageReader* packageReader,
+    MSIX_PACKUNPACK_OPTION packUnpackOptions,
+    MSIX_VALIDATION_OPTION validationOption,
+    char* utf8Destination) noexcept try
+{
+    ThrowErrorIfNot(MSIX::Error::InvalidParameter,
+        utf8Destination != nullptr,
```
msftrubengu: >utf8Destination != nullptr [](start = 8, length = 26)

&& packageReader != nullptr, otherwise we might AV

src/msix/msix.cpp
```diff
@@ -70,6 +70,26 @@ MSIX_API HRESULT STDMETHODCALLTYPE UnpackPackage(
...
+        "Invalid parameters"
+    );
+
+    auto to = MSIX::ComPtr<IStorageObject>::Make<MSIX::DirectoryObject>(utf8Destination);
+
+    MSIX::ComPtr<IAppxPackageReader> reader;
```
msftrubengu: >IAppxPackageReader [](start = 17, length = 18)

You are doing an extra QI
1- QI to IAppxPackageReader
2 -QI to IPackageReader in ComPtr::As

You can make it work with only one QI

src/inc/AppxPackaging.hpp
```diff
@@ -1479,6 +1479,13 @@ MSIX_API HRESULT STDMETHODCALLTYPE UnpackPackage(
     char* utf8Destination
 ) noexcept;
 
+MSIX_API HRESULT STDMETHODCALLTYPE UnpackPackageFromPackageReader(
+    IAppxPackageReader* packageReader,
```
JohnMcPMS: The reader should come after the options for consistency.

src/msix/msix.cpp
```diff
@@ -126,6 +145,30 @@ MSIX_API HRESULT STDMETHODCALLTYPE UnpackBundle(
 #endif
 } CATCH_RETURN();
 
+MSIX_API HRESULT STDMETHODCALLTYPE UnpackBundleFromBundleReader(
+    IAppxBundleReader* bundleReader,
+    MSIX_PACKUNPACK_OPTION packUnpackOptions,
+    MSIX_VALIDATION_OPTION validationOption,
+    MSIX_APPLICABILITY_OPTIONS applicabilityOptions,
```
JohnMcPMS: Not needed here

src/msix/msix.cpp
```diff
@@ -70,6 +70,25 @@ MSIX_API HRESULT STDMETHODCALLTYPE UnpackPackage(
     return static_cast<HRESULT>(MSIX::Error::OK);
 } CATCH_RETURN();
 
+MSIX_API HRESULT STDMETHODCALLTYPE UnpackPackageFromPackageReader(
```
JohnMcPMS: >UnpackPackageFromPackageReader [](start = 35, length = 30)

I would prefer to see the other unpacks call into this one, as it is the natural progression.  But not required in this change.

src/msix/msix.cpp
```diff
@@ -70,6 +70,25 @@ MSIX_API HRESULT STDMETHODCALLTYPE UnpackPackage(
...
+{
+    ThrowErrorIfNot(MSIX::Error::InvalidParameter,
+        (packageReader != nullptr && utf8Destination != nullptr),
+        "Invalid parameters"
+    );
+
+    auto to = MSIX::ComPtr<IStorageObject>::Make<MSIX::DirectoryObject>(utf8Destination);
+
+    MSIX::ComPtr<IPackage> reader;
+    ThrowHrIfFailed(packageReader->QueryInterface(UuidOfImpl<IPackage>::iid, reinterpret_cast<void**>(&reader)));
```
JohnMcPMS: >QueryInterface [](start = 35, length = 14)

@rubengu@microsoft.com , do we not have a "From" on ComPtr?

src/msix/msix.cpp
stephenk-msft: Good idea. I'll go ahead and make this change since it should simplify test validation (won't need to write new tests for the new APIs. The existing tests will cover the new API since they'll eventually call into the new APIs)

src/msix/msix.cpp
msftrubengu: noup :)

---
In reply to: [291764778](https://github.com/Microsoft/msix-packaging/pull/152#discussion_r291764778) [](ancestors = 291764778)

src/inc/AppxPackaging.hpp
JohnMcPMS: But before the destination... for consistency.

---
In reply to: [291763404](https://github.com/Microsoft/msix-packaging/pull/152#discussion_r291763404) [](ancestors = 291763404)

stephenk-msft:
Any additional changes needed? Or is this ready to be checked in?

msftrubengu:
> MSIX_API HRESULT STDMETHODCALLTYPE UnpackPackageFromStream(

Why this one (and the unbundle one) didn't call the new one? 

---
Refers to: src/msix/msix.cpp:92 in 4e7d233. [](commit_id = 4e7d2332196fd4f4dd9d9dac1991ead47b0d8f85, deletion_comment = False)

stephenk-msft:
> MSIX_API HRESULT STDMETHODCALLTYPE UnpackPackageFromStream(
> 
> Why this one (and the unbundle one) didn't call the new one?
> 
> Refers to: src/msix/msix.cpp:92 in 4e7d233. [](commit_id = 4e7d233, deletion_comment = False)

Updated *FromStream APIs! Let me know if they look good

PR #162, opened 2019-06-26
MSIX Installer Istream interface passes installs to win 10 package manager api's

jyvenugo:
- On OS>= 1709, Istream content is written to a temp file and package path is passed to Windows10Redirector::AddPackage.
- Added test in MsixCoreTest.cpp to verify that scribble app installs.
- Removed certain constants previously added for firewall deh which is not being used.


preview/MsixCore/MsixCoreInstallerLib/PackageManager.cpp
```diff
@@ -88,6 +80,70 @@ HRESULT PackageManager::AddPackage(const wstring & packageFilePath, DeploymentOp
...
+                break;
+            }
+        }
+
+        if (!CloseHandle(tempFileHandle))
+        {
+            return HRESULT_FROM_WIN32(GetLastError());
+        }
+
+        RETURN_IF_FAILED(Windows10Redirector::AddPackage(tempFileName));
```
wcheng-msft: move this into a function inside windows10redirector

jyvenugo: Refactored to method in windows10redirector.

PR #170, opened 2019-07-10
IAppxPackageReader tests

msftrubengu:
Adding tests for the AppxPackaging interfaces related to packing.

Fix bugs found while doing the tests:
- Allow/disallow file names.
- Validate that a correct compression option is passed by the caller.
- [Content_Type].xml using Default instead of Override elements.
- Allow creating files with long names for windows

Other changes:
- Simplify AppxPackageWriter::ProcessFileAndAddToPackage signature to take an enum.
- Add std::exception::what to the MSIX Logs
- Move IAppxManifestReader, IAppxBundleManifestReader and IAppxBlockMapReader api tests

TODO:
- E2E tests for PackPackage entrypoint


src/inc/StringStream.hpp
```diff
@@ -64,6 +64,8 @@ namespace MSIX {
             return static_cast<HRESULT>(Error::OK);
         } CATCH_RETURN();
 
+        void Print() { /*std::cout << m_data.str() << std::endl;*/ }
```
JohnMcPMS: >void Print() { /*std::cout << m_data.str() << std::endl;*/ } [](start = 8, length = 60)

Remove?

src/msix/common/FileNameValidation.cpp
```diff
@@ -0,0 +1,307 @@
...
+#include "StringHelper.hpp"
+#include "UnicodeConversion.hpp"
+
+#include <array>
+
+namespace MSIX {
+
+    namespace
+    {
+        constexpr std::size_t ReservedTableSize = 0x7F;
```
JohnMcPMS: >ReservedTableSize [](start = 30, length = 17)

Inconsistent with below definition.

Just use IsReservedFileNameChar.size()

src/msix/pack/AppxPackageWriter.cpp
```diff
@@ -149,22 +148,35 @@ namespace MSIX {
...
-            ProcessFileAndAddToPackage(payloadFiles[i].fileName, stream, payloadFiles[i].compressionOption, payloadFiles[i].contentType, false);
+            ProcessFileAndAddToPackage(payloadFiles[i].fileName, stream, payloadFiles[i].compressionOption, payloadFiles[i].contentType, FileType::AppxPayload);
         }
         failState.release();
         return static_cast<HRESULT>(Error::OK);
     } CATCH_RETURN();
 
     void AppxPackageWriter::ProcessFileAndAddToPackage(const std::string& name, const ComPtr<IStream>& stream,
-        APPX_COMPRESSION_OPTION compressionOpt, const char* contentType, bool forceContentTypeOverride, bool addToBlockMap)
+        APPX_COMPRESSION_OPTION compressionOpt, const char* contentType, FileType fileType)
```
JohnMcPMS: >FileType fileType [](start = 73, length = 17)

This feels more cumbersome than before, although I get that it added some new information you needed.

Could we instead use an overload strategy?

src/msix/pack/XmlWriter.cpp
```diff
@@ -81,18 +81,57 @@ namespace MSIX {
...
             Write("=\"");
-            Write(value);
+            WriteTextValue(value);
             Write("\"");
         }
 
         ComPtr<IStream> XmlWriter::GetStream()
         {
             ThrowErrorIf(Error::XmlError, m_state != State::Finish, "Invalid call, the stream can only be accessed when the writer is done");
+            static_cast<StringStream*>(m_stream.Get())->Print();
```
JohnMcPMS: >static_cast<StringStream*>(m_stream.Get())->Print(); [](start = 12, length = 52)

Remove

src/msix/PAL/FileSystem/Win32/DirectoryObject.cpp
```diff
@@ -116,6 +116,13 @@ namespace MSIX {
     // IDirectoryObject
     ComPtr<IStream> DirectoryObject::OpenFile(const std::string& fileName, FileStream::Mode mode)
     {
+        // TODO: verify this on Win7
```
JohnMcPMS: >TODO: verify this on Win7 [](start = 11, length = 25)

TODONE?

Or pack only?

src/msix/unpack/AppxPackageObject.cpp
```diff
@@ -84,8 +85,7 @@ namespace MSIX {
             // not when bundle support is off.
             THROW_IF_BUNDLE_NOT_ENABLED
             #ifdef BUNDLE_SUPPORT
-            std::string pathInWindows(APPXBUNDLEMANIFEST_XML);
-            std::replace(pathInWindows.begin(), pathInWindows.end(), '/', '\\');
+            std::string pathInWindows = Helper::toBackLash(APPXBUNDLEMANIFEST_XML);
```
JohnMcPMS: >toBackLash [](start = 48, length = 10)

typo

Lash => Slash

src/test/msixtest/inc/PackTestData.hpp
```diff
@@ -0,0 +1,191 @@
...
+        #ifdef WIN32
+        constexpr wchar_t IncompleteHighSurrogate3[] = { L'x', 0xdbff, L'\\', L'x', 0 };
+        constexpr wchar_t UnexpectedLowSurrogate3[] = { 0xd800, 0xdc00, L'\\', 0xdfff, 0 };
+        #else
+        constexpr wchar_t IncompleteHighSurrogate3[] = { L'x', 0xdbff, L'/', L'x', 0 };
+        constexpr wchar_t UnexpectedLowSurrogate3[] = { 0xd800, 0xdc00, L'/', 0xdfff, 0 };
+        #endif
+
+        #ifdef WIN32
+        // File names that should be vaild for adding to a package
```
JohnMcPMS: >vaild [](start = 37, length = 5)

valid

src/test/msixtest/api_packagewriter.cpp
```diff
@@ -0,0 +1,352 @@
...
+{
+    auto outputStream = MsixTest::StreamFile("test_package.msix", false, true);
+
+    MsixTest::ComPtr<IAppxPackageWriter> packageWriter;
+    InitializePackageWriter(outputStream.Get(), &packageWriter);
+
+    // These values are set so that the files added to the package have increasingly
+    // larger sizes, with the first file having a small size < DefaultBlockSize, and
+    // the last file having a large size > 10x DefaultBlockSize.
+    const std::uint32_t contentSizeIncrement = DefaultBlockSize * 10 / static_cast<uint32_t>(TestConstants::GoodFileNames.size()) + 1;
```
JohnMcPMS: >const std::uint32_t contentSizeIncrement = DefaultBlockSize * 10 / static_cast<uint32_t>(TestConstants::GoodFileNames.size()) + 1; [](start = 4, length = 130)

This is a strangely arbitrary value that can change over time. Why?

src/test/msixtest/api_packagewriter.cpp
```diff
@@ -0,0 +1,352 @@
...
+#include "msixtest_int.hpp"
+#include "FileHelpers.hpp"
+#include "PackTestData.hpp"
+#include "macros.hpp"
+
+#include <iostream>
+
+using namespace MsixTest::Pack;
+
+constexpr std::uint32_t DefaultBlockSize = 65535;
```
JohnMcPMS: >DefaultBlockSize [](start = 24, length = 16)

Confusing, as the actual value is 65536...

src/test/msixtest/main.cpp
```diff
@@ -17,9 +17,11 @@ int main( int argc, char* argv[] )
...
     auto find = exePath.find_last_of('/');
     #endif
+    const char* path = nullptr;
     if (find != std::string::npos)
     {
         exePath = exePath.substr(0, find);
+        path = exePath.c_str();
     }
-    return msixtest_main(argc, argv, exePath.c_str());
+    return msixtest_main(argc, argv, path);
```
JohnMcPMS: >path [](start = 37, length = 4)

Why the change here?

msftrubengu: I forgot to add this to the description. This is the bug that if you run the tests from the location where the exe is located, the test will fail to find the location of the test data. 

This is because instead of passing the folder location of the exe, it will pass the exe itself, so it will try to find
(eg. msixtest.exe\testdata\...) which clearly doesn't exists. If we pass nullptr, then the TestPath object will not add the root.


---
In reply to: [302265031](https://github.com/Microsoft/msix-packaging/pull/170#discussion_r302265031) [](ancestors = 302265031)

JohnMcPMS: I see the difference now.

---
In reply to: [302272977](https://github.com/Microsoft/msix-packaging/pull/170#discussion_r302272977) [](ancestors = 302272977,302265031)

src/msix/PAL/FileSystem/Win32/DirectoryObject.cpp
msftrubengu: TODONE and this is actually unpack only.

---
In reply to: [302261077](https://github.com/Microsoft/msix-packaging/pull/170#discussion_r302261077) [](ancestors = 302261077)

src/test/msixtest/api_packagewriter.cpp
msftrubengu: Following the windows tests. I would guess that it was to guarantee that no matter how many filenames are added the last one will be10xDefaultBlockSize-ish

---
In reply to: [302264428](https://github.com/Microsoft/msix-packaging/pull/170#discussion_r302264428) [](ancestors = 302264428)

PR #171, opened 2019-07-10
General improvements to pack

JohnMcPMS:
This change generally improves packing, as well as making a slightly more sweeping change to the zip serialization code.  Changes include:

- In debug builds on Windows, don't assert unless the debugger is attached
- Usage changes to FieldBase, and addition of OptionalFieldBase
  - Additionally, writing the bytes out no longer requires allocation of hundreds of tiny vectors
- Changing stream usage for writing to be more consistent with usage for reading
  - This means instead of ZipFileStream -> RangeStream -> DeflateStream -> VectorStream on RangeStream's vector, we have DeflateStream -> ZipFileStream -> RangeStream -> output stream
  - This also means that we avoid holding entire zip'd streams in memory, instead writing directly to output as we go
- Changed Zip64ExtendedInformation to support the fact that the 4 fields are optional
  - This includes making the entire structure optional, in the event that no field is necessary
- IZipWriter interface changes to reflect the write through nature of the above stream restructuring
- Detection of Zip64 format now based on any of the EoCD fields being -1, rather than just one of them
- Data descriptors for LFH are also now optional, but to match Windows behavior are always generated when not signing

### DIFF
--- a/src/inc/RangeStream.hpp
+++ b/src/inc/RangeStream.hpp
@@ -6,9 +6,7 @@
 #include "Exceptions.hpp"
 #include "StreamBase.hpp"
 #include "ComHelper.hpp"
-#include "VectorStream.hpp"
 #include "MsixFeatureSelector.hpp"
-#include "DeflateStream.hpp"
 
 #include <string>
 #include <map>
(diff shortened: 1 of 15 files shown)


src/inc/RangeStream.hpp
```diff
@@ -78,12 +91,15 @@ namespace MSIX {
...
-            m_size = static_cast<ULONG>(m_buffer.size());
-            m_relativePosition = m_size;
-            // VectorStream will validate if the written bytes are correct. 
-            // It is expected that countBytes != bytesWritten here because this can be a compression.
+            LARGE_INTEGER offset = { 0 };
+            offset.QuadPart = m_relativePosition + m_offset;
+            ThrowHrIfFailed(m_stream->Seek(offset, StreamBase::START, nullptr));
+            ULONG amountWritten = 0;
+            ThrowHrIfFailed(m_stream->Write(buffer, countBytes, &amountWritten));
+            ThrowErrorIf(Error::FileWrite, (countBytes != amountWritten), "Did not write as much as requesteed.");
```
msftrubengu: >requesteed [](start = 100, length = 10)

I think you just copied my or Phil's typo...
Can you change the one in Read too?

PR #172, opened 2019-07-11
PackPackage entrypoint tests

msftrubengu:
Adding tests for PackPackage

- Pack_Good: packs a valid msix package, then unpacks and verifies the extracted files are what is expected.
- Pack_AppxManifestNotPresent: verifies packing a directory without AppxManifest.xml fails
- Pack_AppxManifestWrongLocation: verifies packing a directory without AppxManifest.xml in the top level directory fails.
- Pack_DirectoryNotPresent: verifies packaging a non-existence directory fails.
- Pack_InvalidParameters: verifies passing nullptr as input directory or output package name fails.
- Pack_DeleteFileOnFailure: verifies the output package file is deleted if packaging fails. We only create the stream after the AppxManifest.xml is found, so previous tests can't be used to verify it.


src/test/msixtest/pack.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+    auto testData = MsixTest::TestPath::GetInstance();
+
+    auto directoryPath = testData->GetPath(MsixTest::TestPath::Directory::Pack) + "/" + directory;
+    directoryPath = MsixTest::Directory::PathAsCurrentPlatform(directoryPath);
+
+    auto outputDir = testData->GetPath(MsixTest::TestPath::Directory::Output);
+
+    HRESULT actual = PackPackage(MSIX_PACKUNPACK_OPTION::MSIX_PACKUNPACK_OPTION_NONE,
+                                 MSIX_VALIDATION_OPTION::MSIX_VALIDATION_OPTION_SKIPSIGNATURE,
```
JohnMcPMS: >MSIX_VALIDATION_OPTION_SKIPSIGNATURE [](start = 57, length = 36)

This should have no effect, correct?

src/test/msixtest/pack.cpp
```diff
@@ -0,0 +1,115 @@
...
+// Fails if AppxManifest.xml is not in the top level of the directory
+TEST_CASE("Pack_AppxManifestWrongLocation", "[pack]")
+{
+    HRESULT expected      = static_cast<HRESULT>(MSIX::Error::FileOpen);
+    std::string directory = "input_misplace_manifest";
+
+    RunPackTest(expected, directory);
+}
+
+// Fail if input directory exists
```
JohnMcPMS: >exists [](start = 27, length = 6)

does not exist

src/test/msixtest/pack.cpp
msftrubengu: Correct, at least not for now...

PR #177, opened 2019-07-16
Deprecate MSIX_VALIDATION_OPTION_SKIPAPPXMANIFEST and improve CMake MSIX SDK option processing

msftrubengu:
- Deprecate MSIX_VALIDATION_OPTION_SKIPAPPXMANIFEST by renaming it to MSIX_VALIDATION_OPTION_SKIPAPPXMANIFEST_DEPRECATED. AppxManifest.xml must be always be valid. If the SDK is compiled without USE_VALIDATION_PARSER, no schema validation is done, but it needs to be valid xml. It is recommended to only turn off validation parser for mobile platforms to reduce the size of the binary.
- Move all the MSIX SDK CMake options validation into msix_options.cmake for ease of read. This file is meant to only handle options of the SDK, not build platform specific requirements.
- Always turn on validation parser if --pack is specified in the helper scripts. For MacOS, set xerces as the xml parser and the zlib version that we include in the project instead of forcing adding all the required flags.


CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ option(USE_SHARED_ZLIB "Choose the type of dependency for zlib, Use the -DUSE_SH
 option(USE_STATIC_MSVC "Windows only. Pass /MT as a compiler flag to use the staic version of the run-time library. Default is 'off' (dynamic)" OFF)
 option(SKIP_BUNDLES "Removes bundle functionality from the MSIX SDK. Default is 'off'" OFF)
 option(MSIX_PACK "Include packaging features for the MSIX SDK. Not supported for mobile. Default is 'off'" OFF)
+option(USE_MSIX_SDK_ZLIB "Use zlib implementation under lib/zlib. If off, uses inbox compression library. For Windows and Linux this is no-opt." OFF)
```
JohnMcPMS: Can we move these to msix_options?

cmake/msix_options.cmake
```diff
@@ -0,0 +1,69 @@
...
+        set(CRYPTO_LIB crypt32 CACHE STRING "Crypto Lib not defined. Using crypt32" FORCE)
+    else()
+        set(CRYPTO_LIB openssl CACHE STRING "Crypto Lib not defined. Using openssl" FORCE)
+    endif()
+endif()
+
+# Validates PACK options are correct
+if(MSIX_PACK)
+    if (AOSP OR IOS)
+        message(FATAL_ERROR "Packaging is not supported for mobile devices.")
```
JohnMcPMS: >message(FATAL_ERROR "Packaging is not supported for mobile devices.") [](start = 8, length = 69)

Why?

cmake/msix_options.cmake
```diff
@@ -0,0 +1,69 @@
...
+    elseif(MAC)
+        if(NOT USE_MSIX_SDK_ZLIB)
+            message(FATAL_ERROR "Using libCompression APIs and packaging features is not supported.")
```
JohnMcPMS: >message(FATAL_ERROR "Using libCompression APIs and packaging features is not supported.") [](start = 12, length = 89)

Are these actually a tri-state?  Can we detect 'not specified' and default to the right value, then only error if the user actually specified an incompatible option?

cmake/msix_options.cmake
```diff
@@ -0,0 +1,69 @@
...
+        endif()
+        if(NOT (XML_PARSER MATCHES xerces))
+            message(WARNING "Xerces is the only supported parser for MacOS pack.")
+        endif()
+    endif()
+    if(NOT USE_VALIDATION_PARSER)
+        message(STATUS "validation parsing enabled")
+        set(USE_VALIDATION_PARSER OFF)
```
JohnMcPMS: >OFF [](start = 34, length = 3)

ON?

cmake/msix_options.cmake
msftrubengu: That was one of the initial requirements. We can revisit this once we get a customer need

---
In reply to: [304088645](https://github.com/Microsoft/msix-packaging/pull/177#discussion_r304088645) [](ancestors = 304088645)

cmake/msix_options.cmake
msftrubengu: I don't believe there is a not specified, so if the user don't specify this it means USE_MSIX_SDK_ZLIB=off

---
In reply to: [304089459](https://github.com/Microsoft/msix-packaging/pull/177#discussion_r304089459) [](ancestors = 304089459)

cmake/msix_options.cmake
msftrubengu: Yup, but now we will fail here.

---
In reply to: [304091128](https://github.com/Microsoft/msix-packaging/pull/177#discussion_r304091128) [](ancestors = 304091128)

PR #184, opened 2019-07-31
Implement CreateBundleManifestReader

stephenk-msft:
Added implementation for the CreateBundleManifestReader API.

These changes will enable the application of ACLs to package folders, which is new functionality we are adding to msixmgr.exe.

### DIFF
--- a/src/msix/AppxFactory.cpp
+++ b/src/msix/AppxFactory.cpp
@@ -26,11 +26,9 @@ namespace MSIX {
         IAppxPackageReader** packageReader) noexcept try
     {
         ThrowErrorIf(Error::InvalidParameter, (packageReader == nullptr || *packageReader != nullptr), "Invalid parameter");
-        ComPtr<IMsixFactory> self;
-        ThrowHrIfFailed(QueryInterface(UuidOfImpl<IMsixFactory>::iid, reinterpret_cast<void**>(&self)));
         ComPtr<IStream> input(inputStream);
-        auto zip = ComPtr<IStorageObject>::Make<ZipObject>(self.Get(), input);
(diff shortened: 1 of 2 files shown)


src/msix/AppxFactory.cpp
```diff
@@ -95,14 +96,20 @@ namespace MSIX {
...
     } CATCH_RETURN();
 
-    HRESULT STDMETHODCALLTYPE AppxFactory::CreateBundleManifestReader(IStream *inputStream, IAppxBundleManifestReader **manifestReader) noexcept
+    HRESULT STDMETHODCALLTYPE AppxFactory::CreateBundleManifestReader(IStream *inputStream, IAppxBundleManifestReader **manifestReader) noexcept try
     {
         #ifdef BUNDLE_SUPPORT
-            return static_cast<HRESULT>(Error::NotImplemented);
+            ThrowErrorIf(Error::InvalidParameter, (manifestReader == nullptr || *manifestReader != nullptr), "Invalid parameter");
+            ComPtr<IMsixFactory> self;
+            ThrowHrIfFailed(QueryInterface(UuidOfImpl<IMsixFactory>::iid, reinterpret_cast<void**>(&self)));
```
JohnMcPMS: Should not be needed: this is of type AppxFactory, which derives from IMsixFactory.  Unless there are overrides for both AppxFactory* and IMsixFactory* on AppxBundleManifestObject, it should be able to find the correct constructor and cast the type appropriately.

msftrubengu: can you also fix the other places where this pattern is used?

---
In reply to: [309432387](https://github.com/Microsoft/msix-packaging/pull/184#discussion_r309432387) [](ancestors = 309432387)

PR #187, opened 2019-08-02
Implement IAppxManifestReader4 interface

stephenk-msft:


src/inc/AppxManifestObject.hpp
```diff
@@ -215,8 +215,34 @@ namespace MSIX {
...
+        {
+            *isOptionalPackage = !m_mainPackageName.empty();
+            return static_cast<HRESULT>(Error::OK);
+        } CATCH_RETURN();
+
+        HRESULT STDMETHODCALLTYPE GetMainPackageName(LPWSTR* mainPackageName) noexcept override try
+        {
+            ThrowErrorIf(Error::InvalidParameter, (mainPackageName == nullptr || *mainPackageName != nullptr), "bad pointer");
+            std::wstring utf16MainPackageName = utf8_to_wstring(m_mainPackageName);
+            *mainPackageName = &utf16MainPackageName[0];
```
msftrubengu: >*mainPackageName = &utf16MainPackageName[0]; [](start = 12, length = 44)

This is not the way to marshall out a string. See MarshalOutStringUtf8/MarshalOutString

src/inc/AppxPackaging.hpp
```diff
@@ -1044,6 +1061,23 @@ enum tagLOCKTYPE
     };
 #endif 	/* __IAppxBundleManifestPackageInfo_INTERFACE_DEFINED__ */
 
+#ifndef __IAppxManifestOptionalPackageInfo_INTERFACE_DEFINED__
+#define __IAppxManifestOptionalPackageInfo_INTERFACE_DEFINED__
+
+    // {2634847D-5B5D-4FE5-A243-002FF95EDC7E}
+    MSIX_INTERFACE(IAppxManifestOptionalPackageInfo, 0x54cd06c1, 0x268f, 0x40bb, 0x8e, 0xd2, 0x75, 0x7a, 0x9e, 0xba, 0xec, 0x8d); // what are these numbers for?
+    interface IAppxManifestOptionalPackageInfo : public IUnknown
```
msftrubengu: >IAppxManifestOptionalPackageInfo [](start = 14, length = 32)

For completeness, add IAppxManifestOptionalPackageInfoUtf8.

The sdk also provides utf8 variant interfaces for any interface that has wchar_t  as in or out parameters.

The signature is the same but char is used. 

This is mainly to give support for non-windows platforms.

See other IAppx*Utf8 interface for examples

src/inc/AppxManifestObject.hpp
```diff
@@ -215,8 +215,33 @@ namespace MSIX {
         std::string m_name;
     };
 
+    class AppxManifestOptionalPackageInfo final : public ComClass<AppxManifestOptionalPackageInfo, IAppxManifestOptionalPackageInfo>
+    {
+    public:
+        AppxManifestOptionalPackageInfo(IMsixFactory* factory, std::string mainPackageName) :
```
msftrubengu: >std::string [](start = 63, length = 11)

const std::string&

src/inc/AppxManifestObject.hpp
```diff
@@ -215,8 +215,33 @@ namespace MSIX {
...
+    {
+    public:
+        AppxManifestOptionalPackageInfo(IMsixFactory* factory, std::string mainPackageName) :
+            m_mainPackageName(mainPackageName)
+        {}
+
+        // IAppxManifestOptionalPackageInfo
+        HRESULT STDMETHODCALLTYPE GetIsOptionalPackage(BOOL* isOptionalPackage) noexcept override try
+        {
+            *isOptionalPackage = !m_mainPackageName.empty();
```
msftrubengu: >*isOptionalPackage = !m_mainPackageName.empty(); [](start = 12, length = 48)

I am not sure, but there might be some issues with converting bool to BOOL

src/inc/AppxPackaging.hpp
```diff
@@ -645,6 +649,19 @@ enum tagLOCKTYPE
     };
 #endif 	/* __IAppxManifestReader3_INTERFACE_DEFINED__ */
 
+#ifndef __IAppxManifestReader4_INTERFACE_DEFINED__
+#define __IAppxManifestReader4_INTERFACE_DEFINED__
+
+    // {4579BB7C-741D-4161-B5A1-47BD3B78AD9B}
+    MSIX_INTERFACE(IAppxManifestReader4, 0xc43825ab, 0x69b7, 0x400a, 0x97, 0x09, 0xcc, 0x37, 0xf5, 0xa7, 0x2d, 0x24); // what are these numbers for?
```
JohnMcPMS: >// what are these numbers for? [](start = 118, length = 30)

The GUID of the interface.  Yours do not match the comment, please find the right one and update.

src/inc/AppxManifestObject.hpp
JohnMcPMS: Indeed, I would prefer a trinary here to ensure accurate conversion.

---
In reply to: [310279758](https://github.com/Microsoft/msix-packaging/pull/187#discussion_r310279758) [](ancestors = 310279758)

src/msix/AppxManifestObject.cpp
```diff
@@ -368,6 +368,33 @@ namespace MSIX {
...
+        {
+            std::string mainPackageName;
+        };
+        _context context = {};
+
+        // Parse MainPackageDependency elements
+        XmlVisitor visitorMainPackageDependencies(static_cast<void*>(&context), [](void* c, const ComPtr<IXmlElement>& dependencyNode)->bool
+        {
+            _context* context = reinterpret_cast<_context*>(c);
+            if (context->mainPackageName.empty())
```
JohnMcPMS: >if (context->mainPackageName.empty()) [](start = 12, length = 37)

Instead of checking this, you can just return false when you find the first one, ending the for each.

src/inc/AppxManifestObject.hpp
```diff
@@ -215,8 +215,40 @@ namespace MSIX {
...
+            return static_cast<HRESULT>(Error::OK);
+        } CATCH_RETURN();
+
+        HRESULT STDMETHODCALLTYPE GetMainPackageName(LPWSTR* mainPackageName) noexcept override try
+        {
+            ThrowErrorIf(Error::InvalidParameter, (mainPackageName == nullptr || *mainPackageName != nullptr), "bad pointer");
+            return m_factory->MarshalOutString(m_mainPackageName, mainPackageName);
+        } CATCH_RETURN();
+
+        HRESULT STDMETHODCALLTYPE GetMainPackageName(LPSTR* mainPackageName) noexcept override try
```
msftrubengu: >HRESULT [](start = 8, length = 7)

// IAppxManifestOptionalPackageInfoUtf8

src/msix/AppxManifestObject.cpp
```diff
@@ -368,6 +368,30 @@ namespace MSIX {
         return static_cast<HRESULT>(Error::NotImplemented);
     }
 
+    // IAppxManifestReader4
+    HRESULT STDMETHODCALLTYPE AppxManifestObject::GetOptionalPackageInfo(IAppxManifestOptionalPackageInfo **optionalPackageInfo) noexcept try
```
msftrubengu: > AppxManifestObject::GetOptionalPackageInfo [](start = 29, length = 43)

So for tests... luckily (for you) the test got refactored in the packaging branch, so I wouldn't recommend it adding it now unless you want to deal with merge conflicts later. 

Context: Manifest test used to be in api_package.cpp but, in packaging, they are in api_manifestreader.cpp. The new ones take an actual xml file, as opposed to opening a package every time.

We are targeting to pull packaging into master hopefully early next week. Then you can add tests :)

stephenk-msft: Ok sounds good. I will add the tests next week.

src/inc/ComHelper.hpp
```diff
@@ -57,7 +57,7 @@ namespace MSIX {
     {
         bool Matches(REFIID riid, void** ppvObject)
         {
-            if(ChainInterfaces<I0, Chain...>::Includes(riid))
+            if(riid == UuidOfImpl<I0>::iid || ChainInterfaces<I0, Chain...>::Includes(riid))
```
JohnMcPMS: >riid == UuidOfImpl<I0>::iid [](start = 15, length = 27)

I would do this in the 2 type specialization, so it will hold true for every call.

PR #189, opened 2019-08-05
Packaging master

msftrubengu:
Prepare to release by merging master with packaging.
- Fix build break for no bundle support. 
-  src/test/testData/pack/input/AppxManifest.xml size used to be platform specific so on the test we check for the expected size 3251 for Windows (with CRLF) and 3203 for non windows. After the merge, git decided to always use CRLF even on non-Windows platforms for mysterious reasons.


src/test/msixtest/testData/PackTestData.cpp
```diff
@@ -83,13 +83,7 @@ namespace MsixTest { namespace Pack {
...
         {
             { "AppxBlockMap.xml", 1871 },
-            // The file size differs from Windows and Unix systems (even before packing)
-            // It is possible that is related to LF and CRLF.
-            #ifdef WIN32
-            { "AppxManifest.xml", 3251 },
-            #else
-            { "AppxManifest.xml", 3203 },
-            #endif
+            { "AppxManifest.xml", 3251 }, // This file is marked as LF for git. If you see size = 3251, then it changed to CRLF.
```
JohnMcPMS: >3251 [](start = 34, length = 4)

3203

PR #195, opened 2019-08-07
Initial package signing

JohnMcPMS:
This change adds the initial package signing functionality to the SDK.  It currently supports only one entry point within the binary; the C exported function SignPackage.

The current implementation's features are:
- The incoming package can be signed or unsigned; if signed it will simply be re-signed with the new data
- Only PFX format certificate files are supported
- CodeIntegrity catalog is *not* generated

In addition, makemsix has been updated to include the sign command.  The package, certificate, and format can be supplied, with the format inferred from the certificate file name if needed.

Signing has only been implemented under the OpenSSL Crypto PAL.  While one can still compile for Windows with pack enabled and not use OpenSSL, attempts to sign a package will fail with NotImplemented.

The design of the digest hash collection is to create an accumulator with it's own intelligent routing for the data.  This will allow the writer to pass along data as it is processing it, as well as to do so after the fact when signing an existing package.  This change implements the after the fact signing, but all that is required to sign while creating the package is the piping of the signing info into the empty writer.

Additional changes of note:
- Require that target system be little endian.
- Add static ComPtr::From to remove the need for almost all calls to QI.
- Removed StringStream and renamed VectorStream to MemoryStream.  All uses are now of MemoryStream
- Folded ZipObjectReader into ZipObject
- ZipObjectWriter targets a ZipObject, rather than being one


src/inc/ComHelper.hpp
```diff
@@ -154,10 +154,16 @@ namespace MSIX {
         template <class U>
         ComPtr<U> As() const
         {
-            ComPtr<U> out;
-            ThrowHrIfFailed(m_ptr->QueryInterface(UuidOfImpl<U>::iid, reinterpret_cast<void**>(&out)));
+            return ComPtr<U>::From(m_ptr);
+        }
+
+        static ComPtr<T> From(IUnknown* iunk)
```
msftrubengu: >From(IUnknown* iunk) [](start = 25, length = 20)

I thought it was a myth!

src/inc/ContentTypeWriter.hpp
```diff
@@ -17,15 +18,28 @@ namespace MSIX {
...
         void AddContentType(const std::string& name, const std::string& contentType, bool forceOverride = false);
         void Close();
         ComPtr<IStream> GetStream() { return m_xmlWriter.GetStream(); }
 
     protected:
         void AddDefault(const std::string& ext, const std::string& contentType);
         void AddOverride(const std::string& file, const std::string& contentType);
-        
+
+        static std::string GetPartNameSearchString(const std::string&fileName);
```
msftrubengu: >std::string&fileName [](start = 57, length = 20)

missing space

src/inc/ContentTypeWriter.hpp
```diff
@@ -17,15 +18,28 @@ namespace MSIX {
...
+
+        // File extension to MIME value map that are added as default elements
+        // If the extension is already in the map and its content type is different,
+        // AddOverride is called.
```
msftrubengu: I moved this comment updated into ContentTypeWriter.cpp

src/inc/Exceptions.hpp
```diff
@@ -47,9 +53,16 @@ namespace MSIX {
     class Win32Exception final : public Exception
```
msftrubengu: >class Win32Exception final : public Exception [](start = 4, length = 45)

Maybe we could guard Win32Exception and POSIXException like ThrowHrIfFalse and ThrowHrIfPOSIXFailed?

src/inc/Crypto.hpp
```diff
@@ -3,15 +3,58 @@
...
+        inline void Add(const std::vector<std::uint8_t>& buffer)
+        {
+            Add(buffer.data(), buffer.size());
+        }
+
+        // Gets the hash of the data. This is a destructive action; the accumulated hash
+        // value will be returned and the object can no longer be used.
+        void Get(HashBuffer& hash);
+
+        inline HashBuffer SHA256::Get()
```
msftrubengu: >SHA256:: [](start = 26, length = 8)

Do we need it? I feel it looks weird

src/inc/Exceptions.hpp
JohnMcPMS: We could but they aren't adding any real overhead except symbols.

---
In reply to: [311784039](https://github.com/Microsoft/msix-packaging/pull/195#discussion_r311784039) [](ancestors = 311784039)

src/inc/Crypto.hpp
JohnMcPMS: Copy paste error that actually breaks in clang...

---
In reply to: [311784604](https://github.com/Microsoft/msix-packaging/pull/195#discussion_r311784604) [](ancestors = 311784604)

src/inc/Signing.hpp
```diff
@@ -0,0 +1,144 @@
...
+
+        ~FileAccumulator();
+
+        FileAccumulator(const FileAccumulator&) = delete;
+        FileAccumulator& operator=(const FileAccumulator&) = delete;
+
+        FileAccumulator(FileAccumulator&&) = delete;
+        FileAccumulator& operator=(FileAccumulator&&) = delete;
+
+        inline bool WantsRaw() const { return wantsRaw; }
```
msftrubengu: lol

src/msix/msix.cpp
```diff
@@ -288,4 +289,52 @@ MSIX_API HRESULT STDMETHODCALLTYPE PackPackage(
...
+            signingCertificateFormat == MSIX_CERTIFICATE_FORMAT::MSIX_CERTIFICATE_FORMAT_UNKNOWN,
+            "Certificate format could not be determined");
+
+        ThrowErrorIf(MSIX::Error::InvalidParameter,
+            (MSIX::DoesCertificateFormatRequirePrivateKey(signingCertificateFormat) && privateKey == nullptr),
+            "Certificate format requires separate private key");
+    }
+
+    MSIX::ComPtr<IStream> packageStream = 
+        MSIX::ComPtr<IStream>::Make<MSIX::FileStream>(MSIX::utf8_to_wstring(package).c_str(), MSIX::FileStream::Mode::READ_UPDATE);
```
msftrubengu: >utf8_to_wstring [](start = 60, length = 15)

do we need to use the wstring constructor of FileStream?

JohnMcPMS: Just following the pattern of CreateStreamOnFile

---
In reply to: [315937292](https://github.com/Microsoft/msix-packaging/pull/195#discussion_r315937292) [](ancestors = 315937292)

src/makemsix/main.cpp
```diff
@@ -551,6 +585,51 @@ Command CreatePackCommand()
...
+    };
+
+    result.SetDescription({
+        "Signs the given <package> file, modifying it to either include or replace a",
+        "signature. The <cert> file contains the signing certificate.",
+        "The following certificate formats are supported, and the format will be",
+        "inferred from the file name if not provided:",
+        "    pfx: [*.pfx] A PKCS12 containing both public and private keys",
+        "",
+        "WARNING: This feature is not yet complete. It has only had manual testing",
```
msftrubengu: >WARNING [](start = 9, length = 7)

and it needs to be with openssl

JohnMcPMS:
Postponing adding this to master until we complete it.

PR #196, opened 2019-08-10
Optimize by removing some unnecessary checks

pi1024e:
Did some code cleanup and optimizations here!

### DIFF
--- a/src/msix/PAL/FileSystem/POSIX/DirectoryObject.cpp
+++ b/src/msix/PAL/FileSystem/POSIX/DirectoryObject.cpp
@@ -59,8 +59,8 @@ namespace MSIX {
     void mkdirp(std::string& path, size_t startPos = 0, mode_t mode = DEFAULT_MODE)
     {
         char* p = &path[startPos];
-        if (*p != '/') { p--; }
-        while ( *p != '\0')
+        if (*p == '/') { p++; }
+        while (*p != '\0')
         {
             p++;
             while (*p != '\0' && *p != '/') { p++; }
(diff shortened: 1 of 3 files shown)


src/msix/common/Encoding.cpp
```diff
@@ -301,11 +301,14 @@ namespace MSIX { namespace Encoding {
...
                             }
 
-                            if (!done)
-                            {   // We are not done! Point to the next % encoded UTF-8 seq
-                                sequenceIndex++;
-                                index++;
+                            if (done)
+                            {   
+                                break;
                             }
```
JohnMcPMS: If we are going to "optimize" it, it can currently just be a removal of done and a break where it was set to true.  If this was business logic I would worry about the potential for something to get skipped in the future, but it is unlikely to be the case here.

src/msix/PAL/FileSystem/POSIX/DirectoryObject.cpp
```diff
@@ -60,15 +60,14 @@ namespace MSIX {
     {
         char* p = &path[startPos];
         if (*p == '/') { p++; }
-        while (*p != '\0')
+        for (; *p != '\0'; p++)
```
JohnMcPMS: >for (; *p != '\0'; p++) [](start = 8, length = 23)

Sorry that I missed this before, but this loop will now keep going past the null char.  Previously the termination would be that the inner while loop would reach the null, then create the last directory.  Afterward, the null char would prevent the iteration.

Now, the null char will be reached but the for iteration will be executed every time and push the pointer out past the end of the string.  We will keep mkdir'ing the same existing directory over and over until we reach two nulls in a row to terminate (or SEGSEV).

pi1024e:
Fixed!

pi1024e:
@JohnMcPMS

JohnMcPMS:
@pi1024e ,

Thank you for your contribution, but at this time we cannot accept it as it does not represent a bug fix or a demonstrated and material performance improvement.  I apologize for the inconvenience.

PR #201, opened 2019-08-14
Add ability to unpack and apply acls to package folders in msixmgr.exe

stephenk-msft:
**Adds the following commands to msixmgr.exe:**

1. Unpack: 

Extracts the contents of a package to a folder named according to the package's full name. If the given package is a bundle, each package within the bundle will get unpacked to its own folder. All of the folders will be placed in the parent folder specified by the user through the -destination parameter. The -applyacls option can be used to apply acls to each of the created folders.

`msixmgr.exe -unpack -packagePath <path to package> -destination <parent folder for all created folders> [-applyacls]`

2. ApplyACLs

Applies acls to the given package folder.

`msixmgr.exe -applyacls -packagePath <path to package folder>`

**How Validated:**
- Tested manually
- Passes Windows-internal test automation 
- Unpack logic is covered by existing Azure pipelines tests run when making changes to msix.dll 

**Dependencies**

The new features require msix.dll and applyacls.dll to be in the same directory as msixmgr.exe


preview/MsixCore/msixmgr/msixmgr.cpp
```diff
@@ -137,6 +137,167 @@ int main(int argc, char * argv[])
             std::cout << numPackages << " Package(s) found" << std::endl;
             return S_OK;
         }
+        case OperationType::Unpack:
+        {
+            auto packageFilePath = cli.GetPackageFilePathToInstall();
```
wcheng-msft: i'd recommend putting large chunks of code into separate functions/(possibly files). This can get unwieldy very fast

preview/MsixCore/msixmgr/msixmgr.cpp
```diff
@@ -137,6 +137,167 @@ int main(int argc, char * argv[])
...
+                        packageFolders.push_back(packageFolderName);
+
+                        RETURN_IF_FAILED(packageFilesEnumerator->MoveNext(&hasCurrent));
+                    }
+                }
+
+            }
+            else
+            {
+                return E_INVALIDARG;
```
wcheng-msft: >return [](start = 16, length = 6)

may want to add logging or display something to user.

preview/MsixCore/msixmgr/msixmgr.cpp
```diff
@@ -137,6 +137,167 @@ int main(int argc, char * argv[])
...
+                (GetProcAddress(applyACLsDll, "ApplyACLsToPackageFolder"));
+
+            // Update this to pass folder names to API that apply ACLs to those folders
+            for (auto folder : packageFolders)
+            {
+                std::wcout << folder << std::endl;
+                RETURN_IF_FAILED(ApplyACLsToPackageFolder(folder.c_str()));
+            }
+
+            FreeLibrary(applyACLsDll);
```
wcheng-msft: should this always be done or should it be gated behind IsApplyACLs?

preview/MsixCore/msixmgr/msixmgr.cpp
wcheng-msft: probably 3 functions at least: UnpackPackage() UnpackBundle, and ApplyAcls(), and probaly best to put it in a separate file/class that handles Unpack

---
In reply to: [314106308](https://github.com/microsoft/msix-packaging/pull/201#discussion_r314106308) [](ancestors = 314106308)

preview/MsixCore/msixmgr/msixmgr.cpp
wcheng-msft: >FreeLibrary(applyACLsDll); [](start = 12, length = 26)

you probably want a smart pointer to auto-free on scope exit. As it is the RETURN_IF_FAILED() failure cases above will skip over this.

preview/MsixCore/msixmgr/msixmgr.cpp
```diff
@@ -137,6 +137,167 @@ int main(int argc, char * argv[])
...
+            typedef HRESULT(STDMETHODCALLTYPE *APPLYACLSTOPACKAGEFOLDER)(PCWSTR folderPath);
+
+            APPLYACLSTOPACKAGEFOLDER ApplyACLsToPackageFolder =
+                reinterpret_cast<APPLYACLSTOPACKAGEFOLDER>
+                (GetProcAddress(applyACLsDll, "ApplyACLsToPackageFolder"));
+
+            // Update this to pass folder names to API that apply ACLs to those folders
+            for (auto folder : packageFolders)
+            {
+                std::wcout << folder << std::endl;
```
wcheng-msft: >std::wcout << folder << std::endl; [](start = 16, length = 34)

is this debug output or intended output for user's consumption?

preview/MsixCore/msixmgr/msixmgr.cpp
wcheng-msft: i see that packageFolders would be empty if !IsApplyACLs but you're still loading the applyacls.dll just to do nothing in that case.

---
In reply to: [314111387](https://github.com/microsoft/msix-packaging/pull/201#discussion_r314111387) [](ancestors = 314111387)

preview/MsixCore/msixmgr/msixmgr.cpp
stephenk-msft: This is debug output - should we be logging this to the trace logging provider instead? This could potentially be a lot of output.

preview/MsixCore/msixmgr/msixmgr.cpp
stephenk-msft: You're right, this is a bug. I'll gate this block behind isApplyACLs

preview/MsixCore/msixmgr/msixmgr.cpp
wcheng-msft: yes, log it to the trace logging provider instead

---
In reply to: [314419946](https://github.com/microsoft/msix-packaging/pull/201#discussion_r314419946) [](ancestors = 314419946)

preview/MsixCore/msixmgr/ApplyACLsProvider.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+        typedef HRESULT(STDMETHODCALLTYPE *APPLYACLSTOPACKAGEFOLDER)(PCWSTR folderPath);
+
+        APPLYACLSTOPACKAGEFOLDER ApplyACLsToPackageFolder =
+            reinterpret_cast<APPLYACLSTOPACKAGEFOLDER>
+            (GetProcAddress(*applyACLsDll, "ApplyACLsToPackageFolder"));
+
+        for (auto folder : packageFolders)
+        {
+            std::wcout << folder << std::endl;
```
wcheng-msft: > std::wcout << folder << std::endl; [](start = 11, length = 35)

still here.

preview/MsixCore/msixmgr/ApplyACLsProvider.hpp
```diff
@@ -0,0 +1,21 @@
...
+#include <vector>
+
+namespace MsixCoreLib
+{
+    class ApplyACLsProvider
+    {
+    public:
+
+        ApplyACLsProvider() {};
+        ~ApplyACLsProvider() {};
```
wcheng-msft: you don't need a class. just use a namespace if you have a flat function that does everything. 
My point about class is you could've chosen to integrate the loadlibrary/getprocaddress into the class itself similar to msixmodules.h

preview/MsixCore/msixmgr/UnpackProvider.cpp
```diff
@@ -0,0 +1,170 @@
...
+                std::wstring stdPackageFileName = packageFileName;
+                auto packageFullName = packagesMap[packageFileName];
+                std::wstring packageFolderName = destination + L"\\" + packageFullName;
+                packageFolders.push_back(packageFolderName);
+
+                RETURN_IF_FAILED(packageFilesEnumerator->MoveNext(&hasCurrent));
+            }
+        }
+
+        RETURN_IF_FAILED(ApplyACLsProvider::ApplyACLs(packageFolders));
```
wcheng-msft: >ApplyACLs [](start = 44, length = 9)

shouldn't this be inside the if(isApplyAcls) block abeve

preview/MsixCore/msixmgr/UnpackProvider.cpp
```diff
@@ -0,0 +1,169 @@
...
+        RETURN_IF_FAILED(UnpackPackageFromPackageReader(
+            unpackOption,
+            reader.Get(),
+            &unpackDestination[0]));
+
+        if (isApplyACLs)
+        {
+            std::vector<std::wstring> packageFolders;
+
+            PWSTR packageFullName = nullptr;
```
wcheng-msft: >PWSTR [](start = 12, length = 5)

these need to be released, see Package.cpp for examples on how we release these

preview/MsixCore/msixmgr/ApplyACLsProvider.cpp
wcheng-msft: you haven't addressed this. is it intended for the user to see? you should be able to #include "msixmgrLogger.hpp"  and use TraceLoggingWrite (search in InstallUI.cpp for example)

---
In reply to: [314869606](https://github.com/microsoft/msix-packaging/pull/201#discussion_r314869606) [](ancestors = 314869606)

stephenk-msft: I've still working on validating other parts of the code and was planning to circle back to this later. That being said, do we really want to be logging this to the trace provider? This could literally be hundreds of folder names that are getting logged. This was originally added just so I knew that the correct folders were getting processed

wcheng-msft: oh ok. i  didn't know whether you're waiting on me to review or whether you were doing something else. You definitely don't want it to be cout to user. You can delete if you think there's little value; if there's any value in helping debug in the future, then logging to the trace provider is fine. Trace provider is intended to be noisy--nobody is going to look at it unless something goes wrong.

---
In reply to: [314919298](https://github.com/microsoft/msix-packaging/pull/201#discussion_r314919298) [](ancestors = 314919298)

preview/MsixCore/msixmgr/UnpackProvider.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+        ComPtr<IAppxPackageReader> reader;
+        HRESULT hrCreatePackageReader = factory->CreatePackageReader(stream.Get(), &reader);
+
+        if (FAILED(hrCreatePackageReader))
+        {
+            std::wcout << std::endl;
+            std::wcout << L"Failed with HRESULT " << hrCreatePackageReader << L" when trying to create package reader for " << packageFilePath << L". Please confirm that the certificate for the package has been installed." << std::endl;
+            return hrCreatePackageReader;
+        }
```
wcheng-msft: this should be a specific error code. (or at least your warning about the certificate not being installed should only be displayed when that error code is encountered). I would prefer to do the wcout in msixmgr.cpp  for the specific error code encountered so you don't see multiple error messages when it fails here.

preview/MsixCore/msixmgr/msixmgr.cpp
```diff
@@ -137,6 +139,45 @@ int main(int argc, char * argv[])
...
+                std::wcout << std::endl;
+                std::wcout << "Invalid package path: " << packageFilePath << std::endl;
+                std::wcout << "Please confirm the given package path is an .appx, .appxbundle, .msix, or .msixbundle file" << std::endl;
+                std::wcout << std::endl;
+                return E_INVALIDARG;
+            }
+            if (FAILED(hr))
+            {
+                std::wcout << std::endl;
+                std::wcout << L"Failed with HRESULT " << hr << L" when trying to unpack " << packageFilePath << std::endl;
```
wcheng-msft: >hr [](start = 57, length = 2)

when outputting hr, it's best to have it in hex.

wcheng-msft:
in general, you should fill out the description in the PR that summarizes what the feature is, what is enabled, and how you tested it.

src/test/msixtest/api_manifestreader.cpp
```diff
@@ -623,13 +623,13 @@ TEST_CASE("Api_AppxManifestReader_OptionalPackage", "[api]")
 
     std::vector<std::string> expectedPublisherValues =
     {
-        "", // We expect the value to be NULL. Handle this special case separately  
+        "CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US",
```
msftrubengu: >"CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US", [](start = 8, length = 83)

Just to be sure, you are saying that the previous implementation was wrong and that this one matches the windows implementation?

stephenk-msft: Correct. My original implementation did not match the windows implementation, and this change fixes that.

mobilejon:
Just a question, does msixmgr need to say that it applied ACLS at the end? I noticed when running it that it doesn't say it applied ACLS despite the command being run correctly it just says Finished unpacking packages

stephenk-msft:
Yes, I think that would be helpful to add.

PR #202, opened 2019-08-14
App Execution Alias DEH

jyvenugo:
Parse the manifest for "windows.appExecutionAlias" extension and write related keys to the registry so the app can launch by just specifying the aliasname.
Test: Edited notepad test package to include the alias name and ran through the debugger.
WorkInProgress: The app only launches from Win+R dialog and not the command prompt yet.


jyvenugo:
Closing this PR as I created a new one.

PR #203, opened 2019-08-15
Add a test for packaging a file larger than 4GB

JohnMcPMS:
The primary change is the addition of a test that packs a file larger than 4GB.  This is done by creating a stream that returns easily compressed data, resulting in an ~6MB package file.  The test is tagged with a hidden [.slow] tag, as it takes ~5 minutes to churn through the data and then open the resulting package for verification.

The change also reorganizes the msix in directory into categories for current and future use.  The categories (and directories) are:
- public: headers that we "publish" for consumption by external callers
- shared: headers used by the functional tests; no linking allowed
- internal: headers used by future unit tests, were linking will happen

This allows the tests to reuse product functionality, rather than re-implementing things like ComPtr.


src/test/msixtest/api_packagewriter.cpp
```diff
@@ -350,3 +351,97 @@ TEST_CASE("Api_AppxPackageWriter_closed", "[api]")
...
+        return static_cast<HRESULT>(MSIX::Error::OK);
+    } CATCH_RETURN();
+
+protected:
+    uint64_t m_size = 0;
+    uint64_t m_offset = 0;
+};
+
+// Test creating a valid msix package with a contained file that is larger than 4GB
+// The package itself will be much smaller; do not unpack the package from this test
```
msftrubengu: >do not unpack the package from this test [](start = 44, length = 40)

why? is it just because is going to take a while?

PR #207, opened 2019-08-16
Rename variable to indicate semantics

JohnMcPMS:
Rename the XML interface variable to indicate that it is an xpath query and not just an element name.


src/inc/public/AppxPackaging.hpp
```diff
@@ -1213,7 +1213,7 @@ interface IMsixApplicabilityLanguagesEnumerator;
             /* [retval][string][out] */ LPSTR* value) noexcept = 0;
 
         virtual HRESULT STDMETHODCALLTYPE GetElementsUtf8(
-            /* [in] */ LPCSTR name,
+            /* [in] */ LPCSTR xpath,
```
msftrubengu: >xpath [](start = 30, length = 5)

implementations too? :)

PR #213, opened 2019-08-23
Find package by family name and wildcard support

jyvenugo:
Modified code to enable search by package family name. Also , enabled searching by wildcards(*(match any character) and ?(match single character)). 
When packageFullName or packageFamilyName is passed using wildcards, such as,
msixmgr.exe -FindPackage ***photos***, 
msixmgr.exe -FindPackage *adplus_8wekyb3d8bbw? (PackageFamilyName using wildcards)

it should return the appropriate results.

### DIFF
--- a/preview/MsixCore/msixmgr/msixmgr.cpp
+++ b/preview/MsixCore/msixmgr/msixmgr.cpp
@@ -106,17 +106,17 @@ int main(int argc, char * argv[])
             shared_ptr<IInstalledPackage> packageInfo;
             HRESULT hr = packageManager->FindPackage(cli.GetPackageFullName(), packageInfo);
 
-            if (packageInfo == NULL || hr == ERROR_NOT_FOUND)
+            if (packageInfo == NULL)
             {
                 hr = packageManager->FindPackageByFamilyName(cli.GetPackageFullName(), packageInfo);
                 if (FAILED(hr))
                 {
                     std::wcout << L"Failed to determine findpackage results" << hr << std::endl;
                 }
-                else if (packageInfo == NULL || hr == ERROR_NOT_FOUND)
+                else if (packageInfo == NULL)
                 {
                     std::wcout << std::endl;
-                    std::wcout << L"No packages found " << HRESULT_FROM_WIN32(hr) << std::endl;
+                    std::wcout << L"No packages found " << std::endl;
                     std::wcout << std::endl;
                 }
             }
--- a/preview/MsixCore/msixmgr/msixmgr.cpp
+++ b/preview/MsixCore/msixmgr/msixmgr.cpp
@@ -109,15 +109,15 @@ int main(int argc, char * argv[])
             if (packageInfo == NULL)
             {
                 hr = packageManager->FindPackageByFamilyName(cli.GetPackageFullName(), packageInfo);
-                if (FAILED(hr))
-                {
-                    std::wcout << L"Failed to determine findpackage results" << hr << std::endl;
-                }
-                else if (packageInfo == NULL)
+                if (packageInfo == NULL)
                 {
                     std::wcout << std::endl;
                     std::wcout << L"No packages found " << std::endl;
                     std::wcout << std::endl;
+                }
+                else if (FAILED(hr))
+                {
+                    std::wcout << L"Failed to determine findpackage results" << hr << std::endl;
                 }
             }
 
--- a/preview/MsixCore/msixmgr/msixmgr.cpp
+++ b/preview/MsixCore/msixmgr/msixmgr.cpp
@@ -113,8 +113,7 @@ int main(int argc, char * argv[])
                 {
                     std::wcout << L"Failed to determine findpackage results" << hr << std::endl;
                 }
-
-                if (packageInfo == NULL)
+                else if (packageInfo == NULL)
                 {
                     std::wcout << std::endl;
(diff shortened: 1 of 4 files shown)


preview/MsixCore/msixmgr/msixmgr.cpp
```diff
@@ -105,21 +105,34 @@ int main(int argc, char * argv[])
...
-                std::wcout << std::endl;
-                std::wcout << L"No packages found " << hr << std::endl;
-                std::wcout << std::endl;
+                hr = packageManager->FindPackageByFamilyName(cli.GetPackageFullName(), packageInfo);
+                if (FAILED(hr))
+                {
+                    std::wcout << L"Failed to determine findpackage results" << hr << std::endl;
+                }
+
+                if (packageInfo == NULL)
```
wcheng-msft: >if [](start = 16, length = 2)

else if

jyvenugo: I don't think I can make this else if since package info might be null initially if not found by fullname, but might be populated by the time FindPackageByFamilyName method runs. If 'else if' is introduced, this portion will not execute and results will not be printed even when found from the FindPackageByFamilyName method.

wcheng-msft: this else if would be the else if of the FAILED(hr) above. packageInfo should be null if it fails, and you wouldn't want to log twice.

jyvenugo: Yes, updated that.

wcheng-msft:
add some tests to either (or both) the powershell script or the tests in the test dll.

preview/MsixCore/msixmgrLib/PackageManager.cpp
```diff
@@ -180,20 +198,30 @@ HRESULT PackageManager::FindPackageByFamilyName(const wstring & packageFamilyNam
...
+            if (std::regex_match(installedAppFamilyNameString, packageFamilyNameRegExp))
             {
                 wstring packageDirectoryPath = msixCoreDirectory + std::wstring(p.path().filename());
                 RETURN_IF_FAILED(GetPackageInfo(packageDirectoryPath, installedPackage));
                 return S_OK;
             }
         }
     }
-    return S_OK;
+    return ERROR_NOT_FOUND;
```
wcheng-msft: >ERROR_NOT_FOUND [](start = 11, length = 15)

should be HRESULT_FROM_WIN32(ERROR_NOT_FOUND) as the return type of this should be HRESULT.

preview/MsixCore/msixmgr/msixmgr.cpp
```diff
@@ -105,21 +105,33 @@ int main(int argc, char * argv[])
...
             {
-                std::wcout << std::endl;
-                std::wcout << L"No packages found " << hr << std::endl;
-                std::wcout << std::endl;
+                hr = packageManager->FindPackageByFamilyName(cli.GetPackageFullName(), packageInfo);
+                if (FAILED(hr))
+                {
+                    std::wcout << L"Failed to determine findpackage results" << hr << std::endl;
+                }
+                else if (packageInfo == NULL || hr == ERROR_NOT_FOUND)
```
wcheng-msft: >ERROR_NOT_FOUND [](start = 54, length = 15)

I'd expect ERROR_NOT_FOUND to imply packageInfo == NULL. do you need to separately check this? If you do, this should be HRESULT_FROM_WIN32(ERROR_NOT_FOUND), as FindPackage is supposed to return HRESULT type. 
Also, if you need to check against ERROR_NOT_FOUND explicitly, this should be done before FAILED(hr)

wcheng-msft:
Also, update the readme.md which serves as our documentation for what wildcards we support and maybe give an example too.

preview/MsixCore/msixmgr/msixmgr.cpp
jyvenugo: I do not need to check explicitly for ERROR_NOT_FOUND. packageInfo == NULL should imply the same. I just included that to cover the additional case since the method returns that error. Will remove the check.

preview/MsixCore/msixmgrLib/PackageManager.cpp
jyvenugo: Modified to HRESULT_FROM_WIN32(ERROR_NOT_FOUND).

preview/MsixCore/msixmgr/msixmgr.cpp
jyvenugo: Removed the ERROR_NOT_FOUND check.

jyvenugo: Modified the if else conditions to check for if (hr == HRESULT_FROM_WIN32(ERROR_NOT_FOUND)) first. Since packageInfo would always be NULL even if FAILED(hr), removed the packageInfo==NULL check and replaced it with the ERROR_NOT_FOUND check so that it goes into the else condition too.

jyvenugo:
Added tests to the PowerShell script for FindPackageByFullName, FindPackageByFamilyName, FindPackageByFullName with wildcards, FindPackageByFamilyName with wildcards and FindPackage returns No Packages found cases.

jyvenugo:
Updated ReadMe with explanation and examples.

PR #217, opened 2019-09-13
FindPackage * should return all packages and not just first

jyvenugo:
Addressing the bug where msixmgr.exe -FindPackage * was returning only the first package in the directory. With this change, it should return all results in the directory.


jyvenugo:
-Modified FindPackage and FindPackageByFamilyName methods to not run a regex search since these methods are called by UI and should return exactly one package as per passed package info as input.
-Modified code to handle case insensitive regex search so a* and A* return same results.
-Removed extra code so that in case of no packages found, code just return (0) packages found in same if condition.

wcheng-msft:
code looks fine, do you need to update any tests?

jyvenugo:
> code looks fine, do you need to update any tests?

Added new test package(copy of acdual.msix), Removed FindAllPackages test from PowerShell script, added test to add newly created package as well as acdual.msix, perform -FindPackage a*(testing case insensitive check as well as searching for multiple packages) and checking if both packages are returned.

PR #226, opened 2019-10-01
Fix adding multiple files with the same name via AppxPackaging apis

msftrubengu:
We used to allow adding a file with the same name to the package. The resulting package would seem valid, but it is actually invalid and unpacking will fail, as the blockmap contains multiple entries with the same name. 

This fixes this scenario to fail at packaging time.

### DIFF
--- a/src/test/msixtest/api_packagewriter.cpp
+++ b/src/test/msixtest/api_packagewriter.cpp
@@ -379,7 +379,7 @@ TEST_CASE("Api_AppxPackageWriter_add_same_payload_file")
     // The package should be in an invalid state now.
     REQUIRE_HR(static_cast<HRESULT>(MSIX::Error::InvalidState),
         packageWriter->AddPayloadFile(
-        TestConstants::GoodFileNames[1].second.c_str(),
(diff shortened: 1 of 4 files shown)


src/test/msixtest/api_packagewriter.cpp
```diff
@@ -352,13 +352,46 @@ TEST_CASE("Api_AppxPackageWriter_closed", "[api]")
...
+        packageWriter->AddPayloadFile(
+        TestConstants::GoodFileNames[1].second.c_str(),
+        TestConstants::ContentType.c_str(),
+        APPX_COMPRESSION_OPTION_NORMAL,
+        contentStream.Get()));
+
+    // The package should be in an invalid state now.
+    REQUIRE_HR(static_cast<HRESULT>(MSIX::Error::InvalidState),
+        packageWriter->AddPayloadFile(
+        TestConstants::GoodFileNames[1].second.c_str(),
```
JohnMcPMS: >1 [](start = 37, length = 1)

Make this a different file name just in case.

PR #229, opened 2019-10-03
add VFS file paths and fix caseinsensitiveissubstring

wcheng-msft:
Add some VFS file paths that weren't handled previously, and fix caseInsensitiveSubString which would basically always inadvertently return true.

### DIFF
--- a/preview/MsixCore/msixmgr/FilePaths.cpp
+++ b/preview/MsixCore/msixmgr/FilePaths.cpp
@@ -83,6 +83,7 @@ HRESULT FilePathMappings::InitializePaths()
     TextOle<WCHAR> windowsPath;
     TextOle<WCHAR> commonAppDataPath;
     TextOle<WCHAR> localAppDataPath;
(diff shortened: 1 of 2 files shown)


preview/MsixCore/msixmgr/FilePaths.cpp
```diff
@@ -126,6 +129,8 @@ HRESULT FilePathMappings::InitializePaths()
     m_map[L"AppVSystems32Driverstore"] = appVSystem32DriverstorePath;
     m_map[L"AppVSystems32Logfiles"] = appVSystem32LogfilesPath;
     m_map[L"AppVSystems32Spool"] = appVSystem32SpoolPath;
+    m_map[L"Local AppData"] = std::wstring(localAppDataPath.Get());
```
GITNE: Shouldn't it say `LocalAppData` instead?

wcheng-msft: The AppV VFS path for LocalAppData actually has a space in it.

PR #235, opened 2019-10-10
Load package from programFiles location

jyvenugo:
The package was loading the exe from the VFS location. It should ideally launch from the actual file location. Updated to reflect this.


wcheng-msft:
>         virtual std::wstring GetFullExecutableFilePath()

you can probably just replace this, i don't think this is ever useful

---
Refers to: preview/MsixCore/msixmgr/Package.hpp:140 in 64b79f3. [](commit_id = 64b79f32e3e272ebeb279e8f00a1994de7366537, deletion_comment = False)

jyvenugo:
> virtual std::wstring GetFullExecutableFilePath()
> 
> 
> you can probably just replace this, i don't think this is ever useful
> 
> Refers to: preview/MsixCore/msixmgr/Package.hpp:140 in 64b79f3. [](commit_id = 64b79f3, deletion_comment = False)

Yes, this method was not being used anywhere. Removed it.

PR #238, opened 2019-10-14
Create long file path before passing to mkdirp

jyvenugo:
- With the previous change, certain apps were failing to install because mkdirp was not accepting long file paths. Updated code to append \\\\?\\ before passing to the mkdirp function.
- Updated mkdirp function to start from the 7th character to account for the \\\\?\\.
- Made similar change for populatepackageinfo.cpp for passing long file names.


preview/MsixCore/msixmgrLib/GeneralUtil.cpp
```diff
@@ -246,7 +246,7 @@ namespace MsixCoreLib
     int mkdirp(std::wstring& utf16Path)
```
wcheng-msft: >mkdirp [](start = 8, length = 6)

do all callers of this pass in path with long file prefix? This function should be able to handle both types of path (with/without long file prefix)

jyvenugo: There is one more call to mkdirp from VirtualFileHandler.cpp. Updated that call as well to handle long file paths.

wcheng-msft: fair enough, but usually it's best to have the function handle both prefix'd and not-prefix'd paths; if you prefer not to, then the comments should make it clear to callers that only one format is accepted.

---
In reply to: [334718809](https://github.com/microsoft/msix-packaging/pull/238#discussion_r334718809) [](ancestors = 334718809)

jyvenugo: - Agreed. Updated mkdirp method to handle both prefix'd and not-prefix'd paths rather than letting the method handle only one type of path. 
- Backed out change in VirtualFileHandler.cpp to not pass the path as it was before.

PR #249, opened 2019-10-26
Json Parser for PSF

jyvenugo:
Creating PR to review code changes. Initial parse of json to obtain application values works.
TODO: Match id with that of manifest, organize it better and create class to store values, use values in PSFExecutor.cpp.

### DIFF
--- a/preview/MsixCore/msixmgr/Package.cpp
+++ b/preview/MsixCore/msixmgr/Package.cpp
@@ -174,28 +174,44 @@ HRESULT MsixCoreLib::PackageBase::ProcessPSFIfNecessary()
 
             const Value& apps = doc["applications"];
 
-            for (Value::ConstValueIterator itr =apps.Begin(); itr != apps.End(); ++itr)
+            for (Value::ConstValueIterator itr = apps.Begin(); itr != apps.End(); ++itr)
             {
                 std::string id = (*itr)["id"].GetString();
-                std::string executable = (*itr)["executable"].GetString();
-                std::string arguments = (*itr)["arguments"].GetString();
-                std::string workingDirectory = (*itr)["workingDirectory"].GetString();
+                std::wstring jsonApplicationId;
+                jsonApplicationId.assign(id.begin(), id.end());
+
+                //ApplicationId in json should match applicationId from the xml manifest. Right now, we only process first application from the xml manifest anyway
+                if (CaseInsensitiveEquals(jsonApplicationId, m_applicationId))
+                {
+                    std::string executable = (*itr)["executable"].GetString();
+                    std::wstring psfExecutable;
+                    psfExecutable.assign(executable.begin(), executable.end());
+
+                    std::string arguments = (*itr)["arguments"].GetString();
+                    std::wstring psfArguements;
+                    psfArguements.assign(arguments.begin(), arguments.end());
+
+                    std::string workingDirectory = (*itr)["workingDirectory"].GetString();
+                    std::wstring psfWorkingDirectory;
+                    psfWorkingDirectory.assign(workingDirectory.begin(), workingDirectory.end());
+
+                    // TODO: fill this in with actual stuff. Hardcoding for this xctest package for now
+                    //std::wstring psfExecutable = L"Nokia Siemens Networks\\Managers\\BTS Site\\BTS Site Manager\\jre\\1_6_0\\bin\\javaw.exe";
+                    //std::wstring psfWorkingDirectory = L"Nokia Siemens Networks\\Managers\\BTS Site\\BTS Site Manager";
+                    //m_executionInfo.commandLineArguments = L"-splash:Splash_Wn_BTS_Site_Manager.png -cp cl\\cl.jar -client com.nokia.em.poseidon.PoseidonStarter -confFile cl\\CLConf.xml";
+                    m_executionInfo.commandLineArguments = psfArguements;
+
+                    // resolve the given paths from json into full paths.
+                    m_executionInfo.resolvedExecutableFilePath = FilePathMappings::GetInstance().GetExecutablePath(psfExecutable, m_packageFullName.c_str());
+                    m_executionInfo.workingDirectory = FilePathMappings::GetInstance().GetExecutablePath(psfWorkingDirectory, m_packageFullName.c_str());
+
+                    TraceLoggingWrite(g_MsixTraceLoggingProvider,
+                        "PSF redirection",
+                        TraceLoggingWideString(m_executionInfo.resolvedExecutableFilePath.c_str(), "Resolved PSF executable"),
+                        TraceLoggingWideString(m_executionInfo.workingDirectory.c_str(), "WorkingDirectory"),
+                        TraceLoggingWideString(m_executionInfo.commandLineArguments.c_str(), "Arguments"));
+                }
             }
-
-            // TODO: fill this in with actual stuff. Hardcoding for this test package for now
-            std::wstring psfExecutable = L"Nokia Siemens Networks\\Managers\\BTS Site\\BTS Site Manager\\jre\\1_6_0\\bin\\javaw.exe";
-            std::wstring psfWorkingDirectory = L"Nokia Siemens Networks\\Managers\\BTS Site\\BTS Site Manager";
-            m_executionInfo.commandLineArguments = L"-splash:Splash_Wn_BTS_Site_Manager.png -cp cl\\cl.jar -client com.nokia.em.poseidon.PoseidonStarter -confFile cl\\CLConf.xml";
-
-            // resolve the given paths from json into full paths.
-            m_executionInfo.resolvedExecutableFilePath = FilePathMappings::GetInstance().GetExecutablePath(psfExecutable, m_packageFullName.c_str());
-            m_executionInfo.workingDirectory = FilePathMappings::GetInstance().GetExecutablePath(psfWorkingDirectory, m_packageFullName.c_str());
-
-            TraceLoggingWrite(g_MsixTraceLoggingProvider,
-                "PSF redirection",
-                TraceLoggingWideString(m_executionInfo.resolvedExecutableFilePath.c_str(), "Resolved PSF executable"),
-                TraceLoggingWideString(m_executionInfo.workingDirectory.c_str(), "WorkingDirectory"),
-                TraceLoggingWideString(m_executionInfo.commandLineArguments.c_str(), "Arguments"));
         }
     }
 
--- a/preview/MsixCore/msixmgr/Package.cpp
+++ b/preview/MsixCore/msixmgr/Package.cpp
@@ -195,21 +195,27 @@ HRESULT MsixCoreLib::PackageBase::ProcessPSFIfNecessary()
                     std::wstring psfWorkingDirectory;
                     psfWorkingDirectory.assign(workingDirectory.begin(), workingDirectory.end());
 
-                    // TODO: fill this in with actual stuff. Hardcoding for this xctest package for now
-                    //std::wstring psfExecutable = L"Nokia Siemens Networks\\Managers\\BTS Site\\BTS Site Manager\\jre\\1_6_0\\bin\\javaw.exe";
-                    //std::wstring psfWorkingDirectory = L"Nokia Siemens Networks\\Managers\\BTS Site\\BTS Site Manager";
-                    //m_executionInfo.commandLineArguments = L"-splash:Splash_Wn_BTS_Site_Manager.png -cp cl\\cl.jar -client com.nokia.em.poseidon.PoseidonStarter -confFile cl\\CLConf.xml";
                     m_executionInfo.commandLineArguments = psfArguements;
 
                     // resolve the given paths from json into full paths.
                     m_executionInfo.resolvedExecutableFilePath = FilePathMappings::GetInstance().GetExecutablePath(psfExecutable, m_packageFullName.c_str());
                     m_executionInfo.workingDirectory = FilePathMappings::GetInstance().GetExecutablePath(psfWorkingDirectory, m_packageFullName.c_str());
 
+                    const Value& startScript = (*itr)["startScript"];
+                    std::string scriptPath = startScript["scriptPath"].GetString();
+                    std::wstring psfScriptPath;
+                    psfScriptPath.assign(scriptPath.begin(), scriptPath.end());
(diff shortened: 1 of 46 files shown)


preview/MsixCore/msixmgr/Package.cpp
```diff
@@ -154,20 +160,64 @@ HRESULT MsixCoreLib::PackageBase::ProcessPSFIfNecessary()
...
+                    m_executionInfo.resolvedExecutableFilePath = FilePathMappings::GetInstance().GetExecutablePath(psfExecutable, m_packageFullName.c_str());
+                    m_executionInfo.workingDirectory = FilePathMappings::GetInstance().GetExecutablePath(psfWorkingDirectory, m_packageFullName.c_str());
+
+                    const Value& startScript = (*itr)["startScript"];
+                    std::string scriptPath = startScript["scriptPath"].GetString();
+                    std::wstring psfScriptPath;
+                    psfScriptPath.assign(scriptPath.begin(), scriptPath.end());
+                    m_scriptSettings.scriptPath = psfScriptPath;
+
+                    bool showWindow = startScript["showWindow"].GetBool();
```
wcheng-msft: >showWindow [](start = 25, length = 10)

this should default to false if not present in the json

wcheng-msft:
>     bool waitForFinish = true;

this is also a json setting (default is true)

---
Refers to: preview/MsixCore/msixmgr/PSFScriptExecuter.cpp:47 in 87504ae. [](commit_id = 87504aede07f24c9b31fe967c81bd490823902c3, deletion_comment = False)

preview/MsixCore/msixmgr/Package.cpp
```diff
@@ -154,20 +160,64 @@ HRESULT MsixCoreLib::PackageBase::ProcessPSFIfNecessary()
...
+                    std::wstring psfWorkingDirectory;
+                    psfWorkingDirectory.assign(workingDirectory.begin(), workingDirectory.end());
+
+                    m_executionInfo.commandLineArguments = psfArguements;
+
+                    // resolve the given paths from json into full paths.
+                    m_executionInfo.resolvedExecutableFilePath = FilePathMappings::GetInstance().GetExecutablePath(psfExecutable, m_packageFullName.c_str());
+                    m_executionInfo.workingDirectory = FilePathMappings::GetInstance().GetExecutablePath(psfWorkingDirectory, m_packageFullName.c_str());
+
+                    const Value& startScript = (*itr)["startScript"];
```
wcheng-msft: >startScript [](start = 55, length = 11)

we also want to parse runOnce value, and endScript. In the case of runOnce=false or if endScript is present, we should fail the install, as we do not currently support those cases.

preview/MsixCore/msixmgr/Package.cpp
```diff
@@ -154,20 +160,64 @@ HRESULT MsixCoreLib::PackageBase::ProcessPSFIfNecessary()
...
+
+                //ApplicationId in json should match applicationId from the xml manifest. Right now, we only process first application from the xml manifest anyway
+                if (CaseInsensitiveEquals(jsonApplicationId, m_applicationId))
+                {
+                    std::string executable = (*itr)["executable"].GetString();
+                    std::wstring psfExecutable;
+                    psfExecutable.assign(executable.begin(), executable.end());
+
+                    std::string arguments = (*itr)["arguments"].GetString();
+                    std::wstring psfArguements;
```
wcheng-msft: >psfArguements [](start = 33, length = 13)

psfArguments  (no 'e')

preview/MsixCore/msixmgr/Package.cpp
jyvenugo: If there is an error in parsing the json, should the install be failed in that case as well?

wcheng-msft: yeah. wouldn't that be caught earlier in doc.HasParseError()? you have a comment there to return an error. also keep in mind that it's ok for startScript to be missing entirely from the json. 

---
In reply to: [340250145](https://github.com/microsoft/msix-packaging/pull/249#discussion_r340250145) [](ancestors = 340250145)

wcheng-msft: we should probably also error out if the appplicationId cannot be found

---
In reply to: [340253280](https://github.com/microsoft/msix-packaging/pull/249#discussion_r340253280) [](ancestors = 340253280,340250145)

preview/MsixCore/msixmgr/Package.cpp
jyvenugo: Updated.

jyvenugo:
> bool waitForFinish = true;
> 
> 
> this is also a json setting (default is true)
> 
> Refers to: preview/MsixCore/msixmgr/PSFScriptExecuter.cpp:47 in 87504ae. [](commit_id = 87504ae, deletion_comment = False)

Modified to add this param to scriptSettings and use this value in PSFScriptExecuter.cpp

preview/MsixCore/msixmgr/Package.cpp
jyvenugo: Updated to check if member exists in json first. If exists, then set true/false as per value in json file, else set all boolean values to false. By default, if name value pair is not present in json, rapidjson defaults the value of that param to false.

preview/MsixCore/msixmgr/Package.cpp
jyvenugo: Added checks for runOnce, endScript, return if error in parsing json, and erroring out(return ELEMENT_NOT_FOUND) if app is not found in the applications array at all.

jyvenugo:
Modified to check only for config.json file instead of enumeration. Added checks in json for optional fields prior to extracting values.

preview/MsixCore/msixmgr/Package.cpp
```diff
@@ -148,30 +155,120 @@ HRESULT MsixCoreLib::PackageBase::ProcessPSFIfNecessary()
...
+                std::wstring jsonApplicationId;
+                jsonApplicationId.assign(id.begin(), id.end());
+
+                //ApplicationId in json should match applicationId from the xml manifest. Right now, we only process first application from the xml manifest anyway
+                if (CaseInsensitiveEquals(jsonApplicationId, m_applicationId))
+                {
+                    // Fail if endScript is present in current implementation as we do not support those cases yet
+                    if ((*itr).HasMember("endScript"))
+                    {
+                        return HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED);
```
wcheng-msft: >return HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED) [](start = 24, length = 46)

for each case where we return this, add a TraceLoggingWrite() that explains the reason so that somebody looking at logs can tell which scenario we don't support

preview/MsixCore/msixmgr/Package.cpp
```diff
@@ -148,30 +155,120 @@ HRESULT MsixCoreLib::PackageBase::ProcessPSFIfNecessary()
...
+                            m_scriptSettings.runOnce = runOnce;
+
+                            if (runOnce == false)
+                            {
+                                return HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED);
+                            }
+                        }
+                        else
+                        {
+                            m_scriptSettings.runOnce = false;
```
wcheng-msft: >false [](start = 55, length = 5)

this one defaults to true.

preview/MsixCore/msixmgr/Package.cpp
```diff
@@ -148,30 +155,120 @@ HRESULT MsixCoreLib::PackageBase::ProcessPSFIfNecessary()
...
+                        }
+
+                        if (startScript.HasMember("waitForScriptToFinish"))
+                        {
+                            bool waitForScriptToFinish = startScript["waitForScriptToFinish"].GetBool();
+                            m_scriptSettings.waitForScriptToFinish = waitForScriptToFinish;
+                        }
+                        else
+                        {
+                            m_scriptSettings.waitForScriptToFinish = false;
```
wcheng-msft: >false [](start = 69, length = 5)

this also defaults to true (see the table [https://github.com/microsoft/MSIX-PackageSupportFramework](https://github.com/microsoft/MSIX-PackageSupportFramework))

preview/MsixCore/msixmgr/Package.hpp
```diff
@@ -31,12 +31,16 @@ namespace MsixCoreLib
 
         /// Note: ExecutionInfo is populated by ProcessPSFIsNecessary() and may not be filled in on initialization.
         ExecutionInfo m_executionInfo;
+
+        /// Struct populated by ProcessPSFIsNecessary() and containing scriptrun information 
+        ScriptSettings m_scriptSettings;
```
wcheng-msft: >m_scriptSettings [](start = 23, length = 16)

 = {}

jyvenugo: Initialized values to NULL, true, false, true and removed else case from package.cpp for the Boolean values.

preview/MsixCore/msixmgr/Package.cpp
jyvenugo: Removed else case since its initialized now. If name-value pair is absent, it will default to respective value.

preview/MsixCore/msixmgr/Package.cpp
jyvenugo: Removed else case since its initialized now. If name-value pair is absent, it will default to respective value.

preview/MsixCore/msixmgr/Package.cpp
jyvenugo: Added tracelogging for runonce = false and presence of endScript.

preview/MsixCore/msixmgr/Package.hpp
wcheng-msft: can these be initialized in IPackage.hpp instead? that way the name of the variable is present and better for readability

---
In reply to: [340916640](https://github.com/microsoft/msix-packaging/pull/249#discussion_r340916640) [](ancestors = 340916640)

preview/MsixCore/msixmgr/Package.cpp
```diff
@@ -148,30 +155,115 @@ HRESULT MsixCoreLib::PackageBase::ProcessPSFIfNecessary()
...
-            // resolve the given paths from json into full paths.
-            m_executionInfo.resolvedExecutableFilePath = FilePathMappings::GetInstance().GetExecutablePath(psfExecutable, m_packageFullName.c_str());
-            m_executionInfo.workingDirectory = FilePathMappings::GetInstance().GetExecutablePath(psfWorkingDirectory, m_packageFullName.c_str());
-
-            TraceLoggingWrite(g_MsixTraceLoggingProvider,
-                "PSF redirection",
-                TraceLoggingWideString(m_executionInfo.resolvedExecutableFilePath.c_str(), "Resolved PSF executable"),
-                TraceLoggingWideString(m_executionInfo.workingDirectory.c_str(), "WorkingDirectory"),
-                TraceLoggingWideString(m_executionInfo.commandLineArguments.c_str(), "Arguments"));
+            return HRESULT_FROM_WIN32(ERROR_NOT_SUPPORTED);
```
wcheng-msft: >return [](start = 12, length = 6)

add a  tracelogging here too

preview/MsixCore/msixmgr/Package.cpp
```diff
@@ -148,30 +155,115 @@ HRESULT MsixCoreLib::PackageBase::ProcessPSFIfNecessary()
...
         }
-    }
+        if (doc.HasMember("applications"))
+        {
+            const Value& apps = doc["applications"];
+
+            for (Value::ConstValueIterator itr = apps.Begin(); itr != apps.End(); ++itr)
+            {
+                std::string id = (*itr)["id"].GetString();
+                std::wstring jsonApplicationId;
```
wcheng-msft: >jsonApplicationId [](start = 29, length = 17)

add a TraceLogging here to show what applications we parsed from json

preview/MsixCore/msixmgr/Package.cpp
```diff
@@ -148,30 +155,115 @@ HRESULT MsixCoreLib::PackageBase::ProcessPSFIfNecessary()
...
         TraceLoggingWideString(GetPackageDirectoryPath().c_str(), "path"));
 
-    // By default the PSF information is in config.json, but could be in a different json file.
-    for (auto& p : std::experimental::filesystem::directory_iterator(GetPackageDirectoryPath()))
+    // PSF config information should be in a file called config.json, so look for that file
+    std::wstring jsonConfigFile = GetPackageDirectoryPath() + psfConfigFile;
+    bool configFileExists = false;
+    RETURN_IF_FAILED(FileExists(jsonConfigFile, configFileExists));
+
+    if (configFileExists)
```
wcheng-msft: >configFileExists [](start = 8, length = 16)

i'd invert this if (!configFileExists) {TraceLogging that the file is missing, and return not supported},

preview/MsixCore/msixmgr/Package.cpp
wcheng-msft: and then another one after the for loop finishes to say that we went through all the applications and couldn't find the ID.

---
In reply to: [340917766](https://github.com/microsoft/msix-packaging/pull/249#discussion_r340917766) [](ancestors = 340917766)

preview/MsixCore/msixmgr/Package.hpp
jyvenugo: Initialized in IPackage.hpp

preview/MsixCore/msixmgr/Package.cpp
jyvenugo: Added Config.json has a document parsing error tracelogging here.

preview/MsixCore/msixmgr/Package.cpp
jyvenugo: Added tracelogging for each app id found and for when none is found as well.

preview/MsixCore/msixmgr/Package.cpp
jyvenugo: Inverted this, returned,
ERROR_FILE_NOT_FOUND - when config,json is missing,
ERROR_NOT_SUPPORTED - when doc has parse error, runonce = false and endScript is present.
ERROR_NOT_FOUND - when no application id matching manifest is found.

PR #256, opened 2019-11-13
Add rcconfig.xml file for mui

jyvenugo:
Added rcconfig.xml file which generates a mui file and a language neutral file locally. This change has been committed in order to receive mui files back from api.


wcheng-msft:
I would suggest putting the file in a different folder; either msixmgr or create a new directory

jyvenugo:
> I would suggest putting the file in a different folder; either msixmgr or create a new directory

I put it in this since I added cgmanifest.json in this directory as well.

PR #258, opened 2019-11-15
Replace placeholders with translated strings

wcheng-msft:
replace placeholders with translated strings.

Modify the folder structure to prepare for automating sending the files to the localization service; also mark with a comment to not localize the [MSIXMGR_PRODUCTS] property.


preview/MsixCore/MsixMgrWix/de-de.wxl
```diff
@@ -1,4 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
 <WixLocalization Culture="de-de" xmlns="http://schemas.microsoft.com/wix/2006/localization">
-	<String Id="OrphanedPackageWarning">PLACEHOLDER Warning - removing this will cause the following app(s) installed by this to be unremovable: [MSIXMGR_PRODUCTS]. These app(s) should be removed before proceeding</String>
+	<String Id="OrphanedPackageWarning">Warnung: Wenn Sie dieses Element entfernen, können die folgenden Apps nicht mehr entfernt werden: [MSIXMGR_PRODUCTS]. Diese Apps sollten entfernt werden, bevor Sie den Vorgang fortsetzen.</String>
```
GITNE: Translation guidelines state that it is best to use the non-personal form in German. The proper translation should read like this:

`Warnung: Wenn dieses Element entfernt wrid, können die folgenden Apps nicht mehr entfernt werden: [MSIXMGR_PRODUCTS]. Diese Apps sollten entfernt werden, bevor der Vorgang fortgesetzt wird.`

wcheng-msft: thanks for the feedback; these were the strings returned by our localization service. I've filed an internal bug with the localization folks to see if they agree with you.

wcheng-msft: > Translation guidelines state that it is best to use the non-personal form in German. The proper translation should read like this:
> Warnung: Wenn dieses Element entfernt wird, können die folgenden Apps nicht mehr entfernt werden: [MSIXMGR_PRODUCTS]. Diese Apps sollten entfernt werden, bevor der Vorgang fortgesetzt wird.

We've integrated this with the localization service in a different way now (previously it was submitted as two strings, now it is one), and the service changed the string they returned. Please take a look at https://github.com/microsoft/msix-packaging/pull/278/files If you have concerns with the new way it's worded, let us know. I personally do not read or speak German and so cannot judge as to its quality, but i can pass any concerns along. It is now:
Warnung: Das Entfernen dieses Produkts führt dazu, dass die folgenden App(s), die dadurch installiert werden, nicht mehr entfernt werden können: [MSIXMGR_PRODUCTS]. Diese App(s) sollten vor dem Fortfahren entfernt werden.

GITNE: Thank you for taking so much care.
The new translation is also quite close to the English source text, however it is overly complex. I have one little nit about the tense in the main sentence; it should use the past tense. So, instead of “werden” it should be “wurden”. I guess this is either just a typo or the translator had some trouble understanding the full semantics of this particular message.

`Warnung: Das Entfernen dieses Produkts führt dazu, dass die folgenden App(s), die dadurch installiert wurden, nicht mehr entfernt werden können: [MSIXMGR_PRODUCTS]. Diese App(s) sollten vor dem Fortfahren entfernt werden.`

I do not want to lecture anybody but personally, I do not like “overusing” the dative case on verbs, like in the second sentence. This is something one should generally try to avoid in German when writing. One may use this form but only if it helps to convey an idea or concept more easily than a grammatically fully correct sentence. This style is often used by learners of German without knowing any better. In other words, native German authors and writers would probably use true nouns instead.

Anyhow, I think it would be best to replace the occurrence of `this` by the `[ProductName]` substitute for better understanding.

`Warning - uninstalling [ProductName] will cause the following app(s) installed by [ProductName] to become uninstallable: [MSIXMGR_PRODUCTS]. These app(s) should be removed before proceeding.`

PR #292, opened 2019-12-11
Remove preview directory

wcheng-msft:
rename -- remove preview directory. Fix up associated project files that used relative paths to remove one directory (..\)

ClickOnceWrapper had a problem, in that its dependson was pre-run instead of pre-build, which means it had to have run successfully already; this was fixed.

update README.md to reference paths without preview.


JohnMcPMS:
Update CODEOWNERS for path change as well.

JohnMcPMS:
Also, the root README seems like it should have a more prominent mention of the MsixCore directory.  At least a link to MsixCore\README.md somewhere near the top.

---
In reply to: [564745029](https://github.com/Microsoft/msix-packaging/pull/292#issuecomment-564745029) [](ancestors = 564745029)

wcheng-msft:
> Update CODEOWNERS for path change as well.

ah, that's why you guys were added. updated.

wcheng-msft:
> Also, the root README seems like it should have a more prominent mention of the MsixCore directory.  At least a link to MsixCore\README.md somewhere near the top.
> 
> In reply to: 564745029 [](ancestors = 564745029)

added a sentence and a link.

PR #328, opened 2020-01-17
Decouple apps installed by this instance from the official instance

rooju:
1. Change location of installed applications.
2. Change ARP entry to use packagefamilyname instead of packagefullname.  The ARP entry represents the application as a whole and not any particular version.  Using packagefullname also leaves us vulnerable to a race between the user trying to uninstall and an update bringing in a different version.


rooju:
Sorry, I created this by accident.  This was supposed to go into my private fork.

PR #348, opened 2020-05-28
Handle relative path from current directory.

### LINKED ISSUE #337: Relative path problems when creating MSIX packages
*Project*
MSIX SDK 

**Describe the bug**
When creating MSIX packages with Makemsix, an error message occurs if you use a dot for the -d parameter in the relative path. Error: 0x8000ffff

**To Reproduce**
.\makemsix.exe pack -p ".\MyEmployees.msix" -d ".\PackageFiles"  --> 
it does not work

.\makemsix.exe pack -p ".\MyEmployees.msix" -d "PackageFiles"  --> 
Everything works. 

**Platform**
Win10

msftrubengu:
Fix #337 

The problem is that GetFullPathNameW doesn't return what we expect when lpFileName is a relative path and we fail because we expect newlength to be one less than length (length includes the null character).

```
DWORD length = GetFullPathNameW(pathWide.c_str(), 0, nullptr, nullptr);
DWORD newlength = GetFullPathNameW(pathWide.c_str(), length, &result[prefixChars], nullptr);
ThrowErrorIf(Error::Unexpected, (length - 1) != newlength, "Result length was unexpected");
```

From [MSDN](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfullpathnamew): " Using relative path names in multithreaded applications or shared library code can yield unpredictable results and is not supported... The exception being when the application is guaranteed to be running in a single thread, for example parsing file names from the command line argument string in the main thread prior to creating any additional threads. Using relative path names in multithreaded applications or shared library code can yield unpredictable results and is not supported."

Based on that the MSIX SDK is single-threaded one option is to loose the restriction to throw if length and newlenght is not equal and I verified the string was resolved correctly as length seemed to always be bigger than newlength + 1. However, it doesn't seem safe and we could end up with a truncated path for length -1 < newlength.

The change relax the length match restriction we have and attempt to retry if first call to GetFullPathName result in a smaller size than the second call .


JohnMcPMS:
I would say that long term, the library should not support relative paths, and makemsix (or the calling process) should be responsible for giving full paths.  This will put the burden of knowing their process on the caller.  But for now this should be fine.

msftrubengu:
Talked offline with John and we decided to relax the restrictions to support relative paths. 

Tested with paths like: ".\output" "..\output" "./output" "msixtest\..\output"

src/msix/PAL/FileSystem/Win32/DirectoryObject.cpp
```diff
@@ -130,12 +131,32 @@ namespace MSIX {
...
+            DWORD retry = 1;
+            DWORD previousLength = 0;
+
+            do
+            {
+                previousLength = newLength + retry;
+                result.resize(prefixChars + previousLength, L' ');
+                newLength = GetFullPathNameW(pathWide.c_str(), previousLength, &result[prefixChars], nullptr);
+                retry++;
+            } while ((previousLength > newLength) && retry <= 10);
```
JohnMcPMS, line 155: >previousLength > newLength [](start = 22, length = 26)

Shouldn't this be prev < new ?  As in, "if new is bigger, we need to try again".

Issue #350, opened 2020-06-03
Rename repository to include "MSIX Core" name


Marv51:
As someone following MSIX for years now, when I was looking for the repository of MSIX Core I was puzzled. Googleing (with Bing) didn't really give proper results. 

I looked at this repo multiple times, without realizing this is where all the MSIX Core action is happening.

Please update at least the Readme to make branding here more clear.

(A note that MSIX Core, refers to the heart of MSIX and not dotnet core might also be helpful to people)

jvintzel:
MSIX Core is a variant of the stack that brings minimal functionality to downlevel devices. The goal of this repo is to open sources all aspects of build MSIX which will include more tooling over time.

Marv51:
Point of this issue was to give feedback that the github repo structure is somewhat confusing and the README does not really explain what is where either.

MSIX consists of:
* the [VS packaging project](https://docs.microsoft.com/en-us/windows/msix/desktop/desktop-to-uwp-packaging-dot-net) (closed source)
* the Win10 [appinstaller](https://www.microsoft.com/en-us/p/app-installer/9nblggh4nns1) (closed source)
* the [MSIX Packaging Tool](https://docs.microsoft.com/en-us/windows/msix/packaging-tool/tool-overview) (closed source)
* the [MSIX Core SDK](https://github.com/Microsoft/msix-packaging/tree/master/MsixCore) (A sub-component of this repo)

After a significant amount of reseach into this, the difference between the MSIX Core SDK and the MSIX SDK are unclear to me. 

The MSIX Packaging APIs are a subset of the Packaging API, the readme says. If I understand correctly the MSIX Packaging APIs are the public interface of the MSIX Core SDK? While the Packaging API is the public interface of the packaging component in Windows 10.

Marv51:
Feel free to close this, but I believe that the somewhat confusing naming is major problem for people

Marv51:
I feel like the change of the readme (#353) makes this clearer.

PR #358, opened 2020-07-21
Add Utils library project

lechacon:
Open source SDKUtils source code from the private Windows repo to MSIX SDK repo so that it is available for the Windows community to collaborate/contribute to. This supports our org-wide goal to open source as much of our code as we can. This will also enable us to take other services open source.


msftrubengu:
I am kind of worry that the main directory is called SDKUtils. I think that seeing SDKUtils in a repository where people finds the MSIX SDK will make them think that it is related to it, even though it is not and we are sticking here because reasons... 

We already have an #350 that says that the structure of the repo is confusing

lechacon:
Understood. Do you have any suggestions for the name? Given the contents of this library I can't think of a better description than utils/utilities. Maybe simply "Utils" so that it doesn't seem as related to the MSIX SDK?

jamespik:
agreed about naming as per thread. Something like Tools could also work.

---
In reply to: [662024211](https://github.com/microsoft/msix-packaging/pull/358#issuecomment-662024211) [](ancestors = 662024211)

jamespik:
ah. luis too fast and github->codeflow intergration is pain


---
In reply to: [662089037](https://github.com/microsoft/msix-packaging/pull/358#issuecomment-662089037) [](ancestors = 662089037,662024211)

tools/utils/README.md
```diff
@@ -0,0 +1,4 @@
+This library contains the source for the Utils like AppxPackaging, AppxPackagingInterop,
+CommandLine, IO, Logger, ProcessRunner, and StandardInput. This library will be
+published as a Nuget package and consumed by sources in other repositories like
+MsixPackagingTool.
```
jamespik, line 4: msixpackagingtool isnt public.

None of the consumers today are public except winget cli i believe.

I would leave this very unspecific

tools/utils/Utils/Settings.StyleCop
```diff
@@ -0,0 +1,37 @@
...
+            <Rules>
+                <Rule Name="ElementsMustBeDocumented">
+                    <RuleSettings>
+                        <BooleanProperty Name="Enabled">False</BooleanProperty>
+                    </RuleSettings>
+                </Rule>
+            </Rules>
+            <AnalyzerSettings />
+        </Analyzer>
+        <Analyzer AnalyzerId="Microsoft.StyleCop.CSharp.NamingRules">
```
jamespik, line 27: >icrosoft.StyleCop. [](start = 31, length = 18)

does this work with public stylecop? does this need to be updated to public stylecop?

tools/utils/Utils/AppxPackaging/AppxBundleMetadata.cs
```diff
@@ -0,0 +1,186 @@
...
+        public string RelativePath { get; private set; }
+
+        /// <summary>
+        /// Gets the reader for this bundle
+        /// </summary>
+        public IAppxBundleReader AppxBundleReader { get; private set; }
+
+        /// <summary>
+        /// Gets the list of all the external appx packages referenced by the bundle, which belong to the same family.
+        /// </summary>
```
jamespik, line 57: We should consider find/replacing these with 'msix'

tools/utils/Utils/CommandLine/CommandLineException.cs
```diff
@@ -0,0 +1,43 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="CommandLineException.cs" company="Microsoft">
+//     Copyright (c) Microsoft. All rights reserved.
+// </copyright>
+//-----------------------------------------------------------------------
+
+// Please use CLIApplication instead, this class is being deprecated.
```
jamespik, line 7: >CLIApplication instead, this class is being deprecated. [](start = 14, length = 55)

is anyone still using this? If no callers, delete.

tools/utils/Utils/IO/EmbeddedResourcesUtils.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// Provides methods for accessing assembly embedded resources
+    /// </summary>
+    public class EmbeddedResourcesUtils
+    {
+        /// <summary>
+        /// Returns the embedded resource as a stream.
+        /// </summary>
+        /// <param name="resourceName">the name of the resource</param>
+        /// <returns>the stream containing the resource</returns>
+        public static Stream GetResrouceAsStream(string resourceName)
```
jamespik, line 23: >esrouceA [](start = 33, length = 8)

we should considering fixing these typos while we are rejiggering all this

tools/utils/Utils/IO/EmbeddedResourcesUtils.cs
```diff
@@ -0,0 +1,54 @@
...
+
+            return assembly.GetManifestResourceStream(resourceFullName);
+        }
+
+        /// <summary>
+        /// Returns the embedded resource as a temporary file.
+        /// </summary>
+        /// <param name="resourceName">the name of the resource</param>
+        /// <returns>the path to the temporary file where the resource was copied</returns>
+        public static string GetResrouceAsFile(string resourceName)
```
jamespik, line 39: >esrouc [](start = 33, length = 6)

same

tools/utils/Utils/Logger/AppxPackagingEventSinkToLogger.cs
```diff
@@ -0,0 +1,92 @@
...
+        //   Dispose(false);
+        // }
+
+        // This code added to correctly implement the disposable pattern.
+        public void Dispose()
+        {
+            // Do not change this code. Put cleanup code in Dispose(bool disposing) below.
+            this.Dispose(true);
+
+            // TODO: uncomment the following line if the finalizer is overridden above.
```
jamespik, line 72: > TODO: uncomment the following line if the finalizer is overridden above [](start = 14, length = 72)

we decided not to do this? Shall we purge? Or is disposing the managed states still something we need/should be doing here

tools/utils/Utils/ProcessRunner/ComparePackageRunner.cs
```diff
@@ -0,0 +1,122 @@
...
+            }
+
+            if (string.IsNullOrEmpty(newPackageOrBlockmapPath))
+            {
+                throw new ArgumentException("New package or blackmap path is null or empty", "newPackageOrBlockmapPath");
+            }
+
+            if (!File.Exists(newPackageOrBlockmapPath))
+            {
+                throw new ArgumentException("New package or blackmap path does not exist", "newPackageOrBlockmapPath");
```
jamespik, line 90: Bad typo we should fix.

tools/utils/Utils/ProcessRunner/PackageEditorRunner.cs
```diff
@@ -0,0 +1,454 @@
...
+
+    /// <summary>
+    ///  The base class for Package Editor tool.
+    /// </summary>
+    public sealed class PackageEditorRunner : SDKToolProcessRunner
+    {
+        /// <summary>
+        /// The prefix of test cloud stream uri.
+        /// </summary>
+        private const string CloudStreamPathPrefix = "http://cloudstreamtest?localpath=";
```
jamespik, line 22: >string CloudStreamPathPrefix = "http://cloudstreamtest?localpath="; [](start = 22, length = 67)

doesn't seem like it should be open sourced

tools/utils/Utils/ProcessRunner/PackageEditorRunner.cs
```diff
@@ -0,0 +1,454 @@
...
+            ref string updatedPackagePath,
+            ref string baselinePackageOrBlockMapPath,
+            out bool isCloudStream)
+        {
+            isCloudStream = !string.IsNullOrWhiteSpace(cloudStreamAssemblyName) ||
+                !string.IsNullOrWhiteSpace(cloudStreamClassTypeName);
+
+            Logger.Log(
+                this.LogProviders,
+                "Is Cloud Stream: {0}, Cloud Stream Assembly Name: {1}, Cloud Stream Class Type Name: {2}",
```
jamespik, line 415: Doesnt seem like it should be open sourcced. Is this test infra?

tools/utils/UtilsTests/ProcessRunnerTests/ProcessRunnerTests.cs
```diff
@@ -0,0 +1,286 @@
...
+        /// Test Initialize.
+        /// </summary>
+        [TestInitialize]
+        public new void TestInitialize()
+        {
+            Log.Comment("Initializing ProcessRunner test.");
+
+            // Copy the tool from share under the current directory for testing.
+            this.testDirectory = TestContext.Properties["TestDeploymentDir"].ToString();
+            this.testDataDirectory = this.testDirectory + "peetutils\\test-data\\MakeAppx\\";
```
jamespik, line 36: probably not ok for open sourcing

jamespik: My preference is collateral lives with the repo.

---
In reply to: [458387769](https://github.com/microsoft/msix-packaging/pull/358#discussion_r458387769) [](ancestors = 458387769)

tools/utils/Utils/CommandLine/CommandLineException.cs
lechacon: Yes, this is used, but not CommandLineUtils. I'll delete that and the comment here.

tools/utils/Utils/Logger/AppxPackagingEventSinkToLogger.cs
lechacon: I believe this is part of the code VS auto-generates to implement the Dispose pattern and is safe to remove since we don't use it.

tools/utils/Utils/Settings.StyleCop
lechacon: I believe this is already public StyleCop.

tools/utils/Utils/AppxPackaging/AppxBundleMetadata.cs
lechacon: It makes sense to me for comments. Do you also want to change class and namespace names?

tools/utils/Utils/ProcessRunner/PackageEditorRunner.cs
lechacon: This is a test hook for another project. I'll leave it for now (although I don't like it).

tools/utils/UtilsTests/ProcessRunnerTests/ProcessRunnerTests.cs
lechacon: Added the test data to the repository and updated their names to the correct location.

lechacon:
Ping?

tools/utils/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+This library includes instrumentation to collect usage and diagnostic (error) data and send it to Microsoft to help improve the products using it.
+
+If you build the library yourself the instrumentation will not be enabled and no data will be sent to Microsoft.
+
+The library respects machine wide privacy settings and users can opt-out on their device, as documented in the Microsoft Windows privacy statement [here](https://support.microsoft.com/en-us/help/4468236/diagnostics-feedback-and-privacy-in-windows-10-microsoft-privacy).
+
+In short to opt-out, go to `Start`, then select `Settings` > `Privacy` > `Diagnostics & feedback`, and select `Basic`.
+
+See the [privacy statement](PRIVACY.md) for more details.
```
jamespik, line 39: i assume at this point this has all been vettered/discussed/etc but just in case you may wanna get manu to review this exact language.

tools/utils/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Testing
+
+TAEF is used as the testing framework. This is available through the Windows Driver Kit.
+
+To run the tests, go to `msix-packaging/tools/utils/` and call `te.exe` on `UtilsTests.dll` produced by the build. E.g.
+```
+cd msix-packaging/tools/utils/
+te.exe UtilsTests/bin/Debug/UtilsTests.dll
+```
```
jamespik, line 27: Note where/how you update the makeappx payload from.

jamespik:
> //-----------------------------------------------------------------------

Is the PR history public?  You may need to purge that... for reasons...

---
Refers to: tools/utils/Utils/Telemetry/Telemetry.cs:1 in 042e244. [](commit_id = 042e244a3a818b6337828f18765c0fc9d4420981, deletion_comment = False)

tools/utils/README.md
mapill-msft: This the same statement we used in Winget-CLI, so it should be good to go.

But do we expect to collect any telemetry from this library in future, by replacing the stub helpers?

lechacon: I don't know of any plans to collect telemetry directly from this library. We will replace the stub helpers when building for apps that depend on the wrappers in this library.

msftrubengu:
Consider modifying the CODEOWNERS file. As it is right now, Ran, John and I will show as the required reviewers.

Issue #355, opened 2020-07-09
[BUG] Makemsix pack creates a broken (empty) package when the given directory contains a .git folder


tnguyen63:
*Project*
MSIX SDK

**Describe the bug**
When running makemsix pack with the directory argument that contains a .git, the generated MSIX package is broken as the original content in the given directory isn't in the package when inspecting the package. 

**To Reproduce**
```
cd path\to\directory\of\contents
git init
makemsix pack -d path\to\directory\of\contents -p Package.msix
```
Use 7-zip to inspect the package and we won't see the original content used to create the package.

**Expected behavior**
All the original files/binaries in the directory should be present in the package.

**Screenshots**
Directory given: 
![image](https://user-images.githubusercontent.com/66039363/87077450-fde51680-c1d7-11ea-9136-c58e80461ac0.png)

Output MSIX package:
![image](https://user-images.githubusercontent.com/66039363/87077503-12291380-c1d8-11ea-8cde-0c199982836c.png)

**Platform**
Windows 10 x64

**Additional context**
N/A

msftrubengu:
I was not able to repro this issue, but I fixed a bug where a package that contains .git cannot be unpacked. Please reopen if you see problem again

PR #360, opened 2020-07-27
Fix [Content Types].xml creation for files without an extension

msftrubengu:
At package time, adding files to the [Content Types].xml where a parent directory contains a "." and the actual file doesn't have an extension incorrectly took that "." from a parent directory as a file extension. For example for ".git/description" produces an Extension element `<Default ContentType="application/octet-stream" Extension="git/description"/>` instead of the correctly Override element `<Override ContentType="application/octet-stream" PartName="/.git/config"/>`

#355 uncovered this issue but incorrectly states that the .git folder was not in the package.

### DIFF
--- a/src/msix/pack/ContentTypeWriter.cpp
+++ b/src/msix/pack/ContentTypeWriter.cpp
@@ -44,11 +44,11 @@ namespace MSIX {
     {
         auto percentageEncodedName = Encoding::EncodeFileName(name);
 
-        auto lastDir = percentageEncodedName;
-        auto lastSlash = lastDir.find_last_of("/");
+        auto filename = percentageEncodedName;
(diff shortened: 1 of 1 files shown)


src/msix/pack/ContentTypeWriter.cpp
```diff
@@ -44,17 +44,24 @@ namespace MSIX {
     {
         auto percentageEncodedName = Encoding::EncodeFileName(name);
 
+        auto lastDir = percentageEncodedName;
```
JohnMcPMS, line 47: >lastDir [](start = 13, length = 7)

Typically this would be called "filename", as it is not a directory but the leaf part, aka the file name.

PR #361, opened 2020-07-27
Fix - Makemsix pack fails when directory given ends with "\"

### LINKED ISSUE #354: [BUG] Makemsix pack fails when directory given ends with "\"
*Project*
MSIX SDK

**Describe the bug**
makemsix (Windows) fails to create a package when given a directory argument that ends with a "\\". I believe this is because the tool automatically adds a "\\" to the path. 

**To Reproduce**
`makemsix pack -d path\to\directory\of\contents\ -p Package.msix`

**Expected behavior**
The MSIX package should be created normally. 

**Screenshots**
image

**Platform**
Windows 10 x64.

**Additional context**
N/A

msftrubengu:
Fixes #354 

Pack always appends the specific path separator to the input directory. Because not having a path separator is required in other places in the code, it was easier to just remove it in the constructor if its present.

### DIFF
--- a/src/msix/PAL/FileSystem/POSIX/DirectoryObject.cpp
+++ b/src/msix/PAL/FileSystem/POSIX/DirectoryObject.cpp
@@ -51,12 +51,6 @@ namespace MSIX
             }
         }
 
-        std::vector<std::string> DirectoryObject::GetFileNames(FileNameOptions)
-        {
-            // TODO: Implement when standing-up the pack side for test validation purposes
-            NOTIMPLEMENTED;
-        }
-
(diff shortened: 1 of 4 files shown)


src/msix/PAL/FileSystem/POSIX/DirectoryObject.cpp
```diff
@@ -76,6 +76,12 @@ namespace MSIX {
 
     DirectoryObject::DirectoryObject(const std::string& root, bool createRootIfNecessary) : m_root(root)
     {
+        auto lastOf = m_root.find_last_of(GetPathSeparator());
+        if (lastOf == m_root.length() - 1)
```
JohnMcPMS, line 80: if (!m_root.empty() && m_root.back() == GetPathSeparator())

PR #369, opened 2020-08-11
Change Utils namespace to match NuGet package name

lechacon:
Changing the namespace for the Utils library from Microsoft.Packaging.Utils to Microsoft.Msix.Utils to match what will be the published NuGet package name.

### DIFF
--- a/tools/utils/Utils/Properties/AssemblyInfo.cs
+++ b/tools/utils/Utils/Properties/AssemblyInfo.cs
@@ -11,11 +11,11 @@ using System.Runtime.InteropServices;
 // General Information about an assembly is controlled through the following 
 // set of attributes. Change these attribute values to modify the information
 // associated with an assembly.
-[assembly: AssemblyTitle("Microsoft.Packaging.Utils.Properties")]
+[assembly: AssemblyTitle("Microsoft.Msix.Utils.Properties")]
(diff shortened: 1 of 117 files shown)


tools/utils/Utils/Properties/AssemblyInfo.cs
```diff
@@ -11,11 +11,11 @@
 // General Information about an assembly is controlled through the following 
 // set of attributes. Change these attribute values to modify the information
 // associated with an assembly.
-[assembly: AssemblyTitle("Microsoft.Packaging.Utils.Properties")]
+[assembly: AssemblyTitle("Microsoft.Msix.Utils.Properties")]
 [assembly: AssemblyDescription("Library for utilities like AppxPackaging, AppxPackagingInterop, CommandLine, IO, Logger, ProcessRunner, and StandardInput.")]
```
lechacon, line 15: Does anybody have any suggestion for the package description? I used what we had in the readme as a placeholder, but it's probably better to change it. Something like "Library containing utility types for MSIX, and interop for the Packaging API", perhaps?

jamespik: your proposal sounds reasonable

---
In reply to: [468280667](https://github.com/microsoft/msix-packaging/pull/369#discussion_r468280667) [](ancestors = 468280667)

PR #389, opened 2020-10-16
Implement flat bundle creation

jyvenugo:
-Modifies makemsix to accept and parse input for flat bundle creation. 
-Added PackBundle method to msix.dll.
-Created AppxBundleWriter.cpp to handle bundle creation.
-Parses the input directory(-p parameter) and performs validations on the input to create a vector<PackageInfo> to write to the AppxBundleManifest.xml file.
-Creates the Appxbundlemanifest.xml file, contenttype.xml and Appxblockmap.xml footprint files.

How Verified: I created the .msixbundle file using the command, 'makemsix.exe bundle -d F:\MakeMsix\Test -p F:\MakeMsix\OP2.msixbundle, signed it, put it in a folder with the appx packages that I used to create the bundle and double clicked to install it. I was able to install and launch the app. Flat bundles created with -p and -d parameters work currently.

TODO: Handle -f parameter to create the externalPackages list and write to the bundle manifest.


lib/zlib/zconf.h
```diff
@@ -1,534 +0,0 @@
-/* zconf.h -- configuration of the zlib compression library
```
msftrubengu, line 1: Don't include this file. In the .gitignore there's comment with a command to ignore this file.

src/inc/public/AppxPackaging.hpp
```diff
@@ -124,6 +125,7 @@ interface IAppxManifestQualifiedResourcesEnumerator;
 interface IAppxManifestQualifiedResource;
 interface IAppxBundleFactory;
 interface IAppxBundleWriter;
+interface IAppxBundleWriter4;
```
msftrubengu, line 128: >IAppxBundleWriter4 [](start = 10, length = 18)

You also need to define the interface for non windows platforms (that's why the build is failing in non windows). Look a AppxPackaging.h which gets generated by AppxPackaging.idl and mimic how IAppxBundleWriter4 is defined in here.

src/inc/public/AppxPackaging.hpp
```diff
@@ -1733,6 +1746,14 @@ MSIX_API HRESULT STDMETHODCALLTYPE PackPackage(
 
 #endif // MSIX_PACK
 
+MSIX_API HRESULT STDMETHODCALLTYPE PackBundle(
+    MSIX_BUNDLE_OPTIONS bundleOptions,
+    char* directoryPath,
+    char* outputBundle,
+    char* mappingFile,
```
msftrubengu, line 1753: directoryPath and mappingFile are in inputs for what is going to get bundled, right? I think this might be a little confusing and you might end up with if directoryPath != null && mappingFile != null then bad input.

I think it will look cleaner for callers if there's PackBundle that takes the directory to bundle and PackBundleFromMappingFile. Internally both will call into the same logic

src/msix/msix.cpp
```diff
@@ -289,3 +294,66 @@ MSIX_API HRESULT STDMETHODCALLTYPE PackPackage(
 } CATCH_RETURN();
 
 #endif // MSIX_PACK
+
+MSIX_API HRESULT STDMETHODCALLTYPE PackBundle(
```
msftrubengu, line 298: >PackBundle [](start = 35, length = 10)

This must be in the PACK #ifdef. That's why some builds are failing in windows.

src/msix/common/AppxFactory.cpp
```diff
@@ -88,7 +88,15 @@ namespace MSIX {
     HRESULT STDMETHODCALLTYPE AppxFactory::CreateBundleWriter(IStream *outputStream, UINT64 bundleVersion, IAppxBundleWriter **bundleWriter) noexcept try
     {
         THROW_IF_BUNDLE_NOT_ENABLED
-        NOTIMPLEMENTED;
+        ThrowErrorIf(Error::InvalidParameter, (outputStream == nullptr || bundleWriter == nullptr || *bundleWriter != nullptr), "Invalid parameter");
+        //#ifdef MSIX_PACK
```
msftrubengu, line 92: >//#ifdef MSIX_PACK  [](start = 8, length = 19)

commenting this will also make it fail for non pack builds

src/msix/pack/AppxPackageWriter.cpp
```diff
@@ -65,11 +105,7 @@ namespace MSIX {
 
     HRESULT STDMETHODCALLTYPE AppxPackageWriter::Close(IStream* manifest) noexcept try
     {
-        ThrowErrorIf(Error::InvalidState, m_state != WriterState::Open, "Invalid package writer state");
-        auto failState = MSIX::scope_exit([this]
-        {
-            this->m_state = WriterState::Failed;
-        });
```
msftrubengu, line 72: This is my bad. We should check for the state but only after checking if its a bundle. Also, something can go wrong while reading the manifest file, so the failState should be released here and not in CloseInternal. Can you take that back CloseInternal? The close of the bundle will need something similar.

msftrubengu:
> CodeFlow for GitHub does not download executable files. If you trust this file and would like to run it, please download it yourself through GitHub.

.gitignore

---
Refers to: Thumbs.db:1 in 01cd9b2. [](commit_id = 01cd9b253818cdf356dd40efc6d7249d4d3917d7, deletion_comment = False)

src/inc/internal/AppxBundleWriter.hpp
```diff
@@ -0,0 +1,90 @@
...
+#include "Unknwn.h"
+#include "Objidl.h"
+class IBundleWriter : public IUnknown
+#endif
+{
+public:
+    virtual void ProcessBundlePayload(const MSIX::ComPtr<IDirectoryObject>& from, bool flatBundle) = 0;
+
+};
+MSIX_INTERFACE(IBundleWriter, 0x32e89da5,0x7cbb,0x4443,0x8c,0xf0,0xb8,0x4e,0xed,0xb5,0x1d,0x0a);
```
wcheng-msft, line 35: >0x32e89da5 [](start = 30, length = 10)

you don't want to copy paste this guid, you should generate a new one

src/inc/internal/AppxBundleWriter.hpp
```diff
@@ -0,0 +1,90 @@
...
+#include "BundleWriterHelper.hpp"
+#include "BundleManifestWriter.hpp"
+#include "AppxPackageInfo.hpp"
+
+#include <map>
+#include <memory>
+#include <future>
+
+// internal interface
+// {32e89da5-7cbb-4443-8cf0-b84eedb51d0a}
```
wcheng-msft, line 22: >32e89da5 [](start = 4, length = 8)

should match the generated guid you specify in MSIX_INTERFACE below

src/inc/internal/AppxPackageWriter.hpp
```diff
@@ -36,7 +37,7 @@ namespace MSIX {
         IAppxPackageWriterUtf8, IAppxPackageWriter3, IAppxPackageWriter3Utf8>
     {
     public:
-        AppxPackageWriter(IMsixFactory* factory, const ComPtr<IZipWriter>& zip);
+        AppxPackageWriter(IMsixFactory* factory, const ComPtr<IZipWriter>& zip);//, bool isBundle);
```
wcheng-msft, line 40: >;//, bool isBundle); [](start = 79, length = 20)

remove commented code

src/msix/pack/AppxBundleWriter.cpp
```diff
@@ -0,0 +1,280 @@
...
+        _In_ IStream* packageStream,
+        _In_ bool isDefaultApplicablePackage)
+    {
+        auto appxFactory = m_factory.As<IAppxFactory>();
+
+        ComPtr<IAppxPackageReader> reader;
+        ThrowHrIfFailed(appxFactory->CreatePackageReader(packageStream, &reader));
+
+        UINT64 packageStreamSize = 0;
+        ThrowHrIfFailed(this->m_bundleWriterHelper.GetStreamSize(packageStream, &packageStreamSize));
```
wcheng-msft, line 130: >ThrowHrIfFailed [](start = 8, length = 15)

when you use this, the functions need to CATCH_RETURN()

src/inc/internal/AppxBundleWriter.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+        void AddFileToPackage(const std::string& name, IStream* stream, bool toCompress,
+            bool addToBlockMap, const char* contentType, bool forceContentTypeOverride = false);
+
+        void ValidateCompressionOption(APPX_COMPRESSION_OPTION compressionOpt);
+
+        void CloseInternal();
+
+        HRESULT AddPackageReferenceInternal(_In_ LPCWSTR fileName, _In_ IStream* packageStream,
+            _In_ bool isDefaultApplicablePackage);
```
msftrubengu, line 79: Usually for internal methods we do more of a c++ approach than the com one. For example this method instead of returning a failed HR it will throw an exception that will be caught later on.

src/inc/internal/AppxBundleWriter.hpp
```diff
@@ -0,0 +1,90 @@
...
+            APPX_COMPRESSION_OPTION compressionOpt, const char* contentType);
+
+        void AddFileToPackage(const std::string& name, IStream* stream, bool toCompress,
+            bool addToBlockMap, const char* contentType, bool forceContentTypeOverride = false);
+
+        void ValidateCompressionOption(APPX_COMPRESSION_OPTION compressionOpt);
+
+        void CloseInternal();
+
+        HRESULT AddPackageReferenceInternal(_In_ LPCWSTR fileName, _In_ IStream* packageStream,
```
msftrubengu, line 78: >_In_ [](start = 44, length = 4)

I'm not 100% sure but I believe this will fail in other platforms as SAL is a windows specific feature

msftrubengu, line 78: >LPCWSTR [](start = 49, length = 7)

Internal for the repro you should try to use char (std::string) and not wchar. This helps avoid the hell of wchar in multiple platforms (in windows wchar is 2 bytes and other platforms 4)

src/inc/internal/BundleManifestWriter.hpp
```diff
@@ -0,0 +1,54 @@
...
+namespace MSIX {
+
+    enum ElementWriterState
+    {
+        Uninitialized = 0,
+        BundleManifestStarted = 1,
+        PackagesAdded = 2,
+        OptionalBundlesAdded = 3,
+        BundleManifestEnded = 4,
+    };
```
msftrubengu, line 23: try using scoped enumerations when possible

src/inc/internal/BundleWriterHelper.hpp
```diff
@@ -0,0 +1,95 @@
...
+            UINT64 size;
+            UINT64 offset;
+            //ComPtr<IAppxManifestQualifiedResourcesEnumerator> resources;
+            ComPtr<IAppxManifestResourcesEnumerator> resources;
+
+            BOOL isDefaultApplicablePackage;
+            ComPtr<IAppxManifestTargetDeviceFamiliesEnumerator> tdfs;
+        };
+
+    class BundleWriterHelper
```
msftrubengu, line 34: >BundleWriterHelper [](start = 10, length = 18)

This class looks a lot like windows code, instead of the general idea of cross-plat application. 

You can change the contract here to don't be hresult based and use exceptions. also that way you can avoid having out parameters and just return the value

src/inc/internal/BundleWriterHelper.hpp
```diff
@@ -0,0 +1,95 @@
...
+            UINT64 version;
+            std::string architecture;
+            std::string resourceId;
+            PCWSTR fileName;
+            UINT64 size;
+            UINT64 offset;
+            //ComPtr<IAppxManifestQualifiedResourcesEnumerator> resources;
+            ComPtr<IAppxManifestResourcesEnumerator> resources;
+
+            BOOL isDefaultApplicablePackage;
```
msftrubengu, line 30: >BOOL [](start = 12, length = 4)

use bool

src/inc/internal/BundleWriterHelper.hpp
```diff
@@ -0,0 +1,95 @@
...
+#include <map>
+#include <memory>
+#include <future>
+
+namespace MSIX {
+
+    struct PackageInfo
+        {
+            APPX_BUNDLE_PAYLOAD_PACKAGE_TYPE type;
+            UINT64 version;
```
msftrubengu, line 21: >UINT64 [](start = 12, length = 6)

try using the std types

src/inc/internal/BundleWriterHelper.hpp
```diff
@@ -0,0 +1,95 @@
...
+            std::string architecture;
+            std::string resourceId;
+            PCWSTR fileName;
```
msftrubengu, line 24: >PCWSTR [](start = 12, length = 6)

this should be a std::string. internally everything should be std::string and only when we marshal out a string the conversion to wstring should be done

src/inc/internal/FileNameValidation.hpp
```diff
@@ -13,7 +15,8 @@ namespace MSIX {
     {
         static bool IsFileNameValid(const std::string& name);
         static bool IsIdentifierValid(const std::string& name);
-        static bool IsFootPrintFile(const std::string& fileName);
+        static bool IsFootPrintFile(const std::string& fileName, bool isBundle);
         static bool IsReservedFolder(const std::string& fileName);
+        //static HRESULT IsUnderStubFolder(const std::string& fileName, bool* IsUnderStubFolder);
```
msftrubengu, line 20: >//static HRESULT IsUnderStubFolder(const std::string& fileName, bool* IsUnderStubFolder); [](start = 8, length = 89)

remove

src/makemsix/main.cpp
```diff
@@ -236,6 +237,7 @@ struct Invocation
         auto option = std::find(options.begin(), options.end(), name);
         return (option == options.end() ? nullptr : &*option);
     }
+
 };
```
msftrubengu, line 241: nit: remove blank line

src/msix/pack/AppxBundleWriter.cpp
```diff
@@ -0,0 +1,280 @@
...
+
+        CloseInternal();
+        return static_cast<HRESULT>(Error::OK);
+    } CATCH_RETURN();
+
+    // IAppxBundleWriter4
+    HRESULT STDMETHODCALLTYPE AppxBundleWriter::AddPackageReference(LPCWSTR fileName, 
+        IStream* inputStream, BOOL isDefaultApplicablePackage) noexcept try
+    {        
+        HRESULT hr = this->AddPackageReferenceInternal(fileName, inputStream, !!isDefaultApplicablePackage);
```
msftrubengu, line 108: >HRESULT hr =  [](start = 8, length = 13)

ThrowHrIfFailed

src/msix/pack/AppxBundleWriter.cpp
```diff
@@ -0,0 +1,280 @@
...
+        bool result = ((compressionOpt == APPX_COMPRESSION_OPTION_NONE) ||
+                       (compressionOpt == APPX_COMPRESSION_OPTION_NORMAL) ||
+                       (compressionOpt == APPX_COMPRESSION_OPTION_MAXIMUM) ||
+                       (compressionOpt == APPX_COMPRESSION_OPTION_FAST) ||
+                       (compressionOpt == APPX_COMPRESSION_OPTION_SUPERFAST));
+        ThrowErrorIfNot(Error::InvalidParameter, result, "Invalid compression option.");
+    }
+
+    // Common close functionality for pack and bundle.
+    // Adds AppxBlockMap.xml, [Content_Types].xml and closes the zip object.
```
msftrubengu, line 255: You can remove this. My idea was to use AppxPackageWriter as both the package and the bundle writer, but since its now separate this comment shouldn't go here

src/msix/pack/AppxPackageWriter.cpp
```diff
@@ -254,4 +238,30 @@ namespace MSIX {
         ThrowErrorIfNot(Error::InvalidParameter, result, "Invalid compression option.");
     }
 
+    // Common close functionality for pack and bundle.
```
msftrubengu, line 241: >// Common close functionality for pack and bundle. [](start = 4, length = 50)

same here, you can actually revert this file entirely and fix the build breaks

src/inc/internal/AppxBundleWriter.hpp
jyvenugo: Generated and added new guid in here.

src/inc/internal/AppxBundleWriter.hpp
jyvenugo: Generated new guid and matched it with the one defined above.

src/inc/internal/AppxPackageWriter.hpp
jyvenugo: Removed.

src/inc/internal/AppxBundleWriter.hpp
jyvenugo: Removed the SAL annotation, _In_ and used std::string instead of LPCWSTR.

jyvenugo: Used std::string instead instead of wchar.

src/inc/internal/BundleManifestWriter.hpp
jyvenugo: Removed this enum since it was not being used anywhere as of now.

src/inc/internal/BundleWriterHelper.hpp
jyvenugo: Removed this field as this is not being used anywhere as of now.

src/inc/internal/BundleWriterHelper.hpp
jyvenugo: Updated to use std::string and std::uint64_t.

src/inc/internal/BundleWriterHelper.hpp
jyvenugo: Updated all the fields to use standard std types.

src/inc/internal/FileNameValidation.hpp
jyvenugo: Removed.

src/msix/pack/AppxBundleWriter.cpp
jyvenugo: Updated to use ThrowHrIfFailed.

src/makemsix/main.cpp
jyvenugo: Removed blank line.

src/msix/pack/AppxPackageWriter.cpp
jyvenugo: Reverted this file entirely since I created a separate file for bundles, AppxBundleWriter.cpp.

src/msix/pack/AppxPackageWriter.cpp
jyvenugo: Reverted this file entirely except for the change to the IsFootPrintFile method in ValidationHelper since that method checks for different files based on if it is a package or a bundle.

lib/zlib/zconf.h
jyvenugo: Reverted change to this file.

src/msix/pack/AppxBundleWriter.cpp
jyvenugo: Removed CloseInternal method and moved it to ::Close in the AppxBundleWriter.cpp file.

src/msix/pack/AppxBundleWriter.cpp
jyvenugo: Updated to return void for all internal methods except methods implementing an interface(which returns HRESULT and uses CATCH_RETURN()

src/inc/internal/AppxBundleWriter.hpp
jyvenugo: Updated to return void for most internal methods and use exceptions except for methods implementing an interface which returns HRESULT.

src/inc/internal/BundleWriterHelper.hpp
jyvenugo: Updated most methods in this file to use exceptions instead of hresults.

src/inc/public/AppxPackaging.hpp
```diff
@@ -1733,6 +1746,14 @@ MSIX_API HRESULT STDMETHODCALLTYPE PackPackage(
 
 #endif // MSIX_PACK
 
+MSIX_API HRESULT STDMETHODCALLTYPE PackBundle(
```
msftrubengu, line 1749: >PackBundle [](start = 35, length = 10)

This should be inside the #ifdef MSIX_PACK

src/inc/internal/AppxManifestObject.hpp
```diff
@@ -317,10 +356,8 @@ namespace MSIX {
         HRESULT STDMETHODCALLTYPE GetApplications(IAppxManifestApplicationsEnumerator **applications) noexcept override;
         HRESULT STDMETHODCALLTYPE GetStream(IStream **manifestStream) noexcept override;
 
-        // IAppxManifestReader2
-        HRESULT STDMETHODCALLTYPE GetQualifiedResources(IAppxManifestQualifiedResourcesEnumerator **resources) noexcept override;
```
msftrubengu, line 321: I think you should keep it like this to be clear that this method comes from this interface, even though you can call this method via IAppxManifestReader3 or IAppxManifestReader4 because of inheritance

src/inc/internal/BundleWriterHelper.hpp
```diff
@@ -0,0 +1,80 @@
...
+#include "AppxBlockMapWriter.hpp"
+#include "ContentTypeWriter.hpp"
+#include "ZipObjectWriter.hpp"
+#include "AppxPackageInfo.hpp"
+#include "BundleManifestWriter.hpp"
+
+#include <map>
+#include <memory>
+#include <future>
+
```
msftrubengu, line 15: Do you use this?

src/inc/internal/GeneralUtil.hpp
```diff
@@ -0,0 +1,34 @@
+#pragma once
```
msftrubengu, line 1: This file was meant to be a helper for callers to the dll. Internally we shouldnt use it and use std::string, we usually implement internal interface to return std::string when needed.

src/makemsix/main.cpp
```diff
@@ -212,6 +212,7 @@ struct Invocation
         return opt->params[0];
     }
 
+
```
msftrubengu, line 215: nit: here too :)

src/makemsix/main.cpp
```diff
@@ -567,6 +659,7 @@ int main(int argc, char* argv[])
         #ifdef MSIX_PACK
         CreatePackCommand(),
         #endif
+        CreateBundleCommand(),
```
msftrubengu, line 662: >CreateBundleCommand(), [](start = 8, length = 22)

This should only be present if MSIX_PACK

src/msix/msix.cpp
```diff
@@ -289,3 +293,110 @@ MSIX_API HRESULT STDMETHODCALLTYPE PackPackage(
...
+
+    if (bundleOptions & MSIX_BUNDLE_OPTIONS::MSIX_OPTION_OVERWRITE)
+    {
+        overWrite = true;
+    }
+
+    if (bundleOptions & MSIX_BUNDLE_OPTIONS::MSIX_OPTION_NOOVERWRITE)
+    {
+        overWrite = false;
+    }
```
msftrubengu, line 329: if (bundleOptions & MSIX_BUNDLE_OPTIONS::MSIX_OPTION_OVERWRITE || bundleOptions & MSIX_BUNDLE_OPTIONS::MSIX_OPTION_NOOVERWRITE)

src/msix/msix.cpp
```diff
@@ -17,7 +18,10 @@
 #include "AppxPackageObject.hpp"
 #include "MsixFeatureSelector.hpp"
 #include "AppxPackageWriter.hpp"
+#include "AppxBundleWriter.hpp"
 #include "ScopeExit.hpp"
+#include "MSIXWindows.hpp"
+#include "GeneralUtil.hpp"
```
msftrubengu, line 24: remove

src/msix/pack/AppxBundleWriter.cpp
```diff
@@ -0,0 +1,240 @@
...
+#include "Encoding.hpp"
+#include "ZipObjectWriter.hpp"
+#include "AppxManifestObject.hpp"
+#include "ScopeExit.hpp"
+#include "FileNameValidation.hpp"
+#include "StringHelper.hpp"
+
+#include <string>
+#include <memory>
+#include <future>
```
msftrubengu, line 20: Same

src/msix/pack/AppxBundleWriter.cpp
```diff
@@ -0,0 +1,240 @@
...
+#include <algorithm>
+#include <functional>
+
+#include <zlib.h>
```
msftrubengu, line 24: Do you need this? I would expect that the zipwriter encapsulates all of the zlib code

src/msix/pack/AppxBundleWriter.cpp
```diff
@@ -0,0 +1,240 @@
...
+namespace MSIX {
+
+    AppxBundleWriter::AppxBundleWriter(IMsixFactory* factory, const ComPtr<IZipWriter>& zip, std::uint64_t bundleVersion)
+        : m_factory(factory), m_zipWriter(zip)
+    {
+        m_state = WriterState::Open;
+        if(bundleVersion == 0)
+        {
+            SYSTEMTIME time = {0};
+            GetSystemTime(&time);
```
msftrubengu, line 35: This won't work for non windows platforms. You need to either define a PAL or to look for c/c++ std functions that gives you this information

src/inc/public/AppxPackaging.hpp
jyvenugo: I added checks for when directoryPath != null && mappingFile != null within this method. Also added checks for other possible scenarios when directoryPath != null && mappingFile != null.

src/inc/internal/GeneralUtil.hpp
jyvenugo: I added this file, which has helper functions needed to process the input in the PackBundle method. It does version conversion to and from uint64, checks if output file is already present or if the output file is a directory. Do you mean I should not be adding this file?

src/msix/msix.cpp
jyvenugo: I removed the MSIXWindows.hpp file but I need the GeneralUtil.hpp file to be there because it has helper functions to process the input in the PackBundle method.

src/inc/internal/BundleWriterHelper.hpp
jyvenugo: I use the map for std::map<std::string, OptionalBundleInfo> optionalBundles;

src/msix/pack/AppxBundleWriter.cpp
jyvenugo: Removed.

src/msix/pack/AppxBundleWriter.cpp
jyvenugo: Build fails without zlib.h. Removed the other two.

src/makemsix/main.cpp
jyvenugo: Removed.

src/inc/internal/AppxManifestObject.hpp
jyvenugo: Updated to reflect this.

src/inc/public/AppxPackaging.hpp
jyvenugo: I defined interface for IAppxBundleWriter4.

src/msix/msix.cpp
jyvenugo: Updated.

src/msix/common/AppxFactory.cpp
jyvenugo: Updated.

src/inc/public/AppxPackaging.hpp
jyvenugo: Updated.

src/makemsix/main.cpp
jyvenugo: Updated.

src/msix/pack/AppxBundleWriter.cpp
jyvenugo: Replaced with std::time_t and formatted suing std::put_time

wcheng-msft (approved):
:shipit:

src/inc/internal/AppxBundleWriter.hpp
```diff
@@ -0,0 +1,82 @@
...
+#include "ContentTypeWriter.hpp"
+#include "ZipObjectWriter.hpp"
+#include "BundleWriterHelper.hpp"
+#include "BundleManifestWriter.hpp"
+#include "AppxPackageInfo.hpp"
+
+#include <map>
+#include <memory>
+#include <future>
+
```
msftrubengu, line 20: Make sure to only include headers when you actually need them

src/inc/internal/GeneralUtil.hpp
msftrubengu: Talk offline about this. The goal should be to never use an RAII wrapper for strings as it requires marshaling out a string and inside the dll boundary that's not the best approach because of performance. Also this class assumes the factory object uses std::malloc  which might not be the case for some windows callers. 

As for ConvertVersionStringToUint64 ConvertVersionToString I think we can rename the file to VersionHelpers.hpp/cpp to reflect more the nature of these methods or perhaps add them in StringHelper.hpp and create a StringHelper.cpp

I think that GeneralUtil is just to board and anything can live here, so separation will make the code more easy to read in the future.

---
In reply to: [517406849](https://github.com/microsoft/msix-packaging/pull/389#discussion_r517406849) [](ancestors = 517406849)

src/msix/pack/AppxBundleWriter.cpp
msftrubengu: Can you move the #include <zlib.h> to ZipWriter.hpp? 

---
In reply to: [516882961](https://github.com/microsoft/msix-packaging/pull/389#discussion_r516882961) [](ancestors = 516882961)

src/inc/internal/AppxBundleWriter.hpp
jyvenugo: Removed these headers.

src/inc/internal/GeneralUtil.hpp
jyvenugo: Removed the extra TEXT structure that was being used. Created IAppxTargetDeviceFamilyInternal and IAppxQualifiedResourcesInternal instead to fetch these string values. Renamed the GeneralUtil file to VersionHelpers. I will create function specific file names going forward.

src/msix/pack/AppxBundleWriter.cpp
jyvenugo: Moved #include <zlib.h> to ZipWriter.cpp.

msftrubengu (approved):
:shipit:

msftrubengu (approved):
:shipit:

PR #391, opened 2020-11-02
Allow users to mount CIM with image path instead of volume id

stephenk-msft:
Currently, the workflows for mounting/unmounting VHD and CIM files using msixmgr.exe is inconsistent. VHDs are mounted via image file path, and CIMs are mounted via a user-specified volume id.

This PR brings the two workflows into alignment by enabling users to mount CIMs by providing just the image file path. Under the hood, CreateCIM.dll will generate the volume id so that the user does not need to specify one. In addition, CreateCIM.dll will store image file path --> volume if mappings in the registry so that unmounting the CIM file can also be performed when provided only the image file path. 

To-Do:
- Upload the updated CreateCIM.dll binary


wcheng-msft (approved):
:shipit:

PR #401, opened 2020-12-29
Installation fails if certificate contains extended ascii or non alpha-numeric character

jyvenugo:
**Issue:** The installation was failing if the app had an accent in the subject name of the certificate or the publisher in the manifest.
**Reason**: 
The equals comparison check between the publisher name derived from the certificate and the publisher name from the appx manifest file was failing. It appears that the value of the 'publisher' being read from the signature file(certificate) was being read and derived using CertNameToStrA which was returning the ANSI string. I changed it to CertNameToStrW which now returns the correct UTF8 string.

**Verified**: by running an app which is signed with a certificate which has the following special characters content in it subject name and manifest publisher name - Île-de-France. It installs and runs successfully.

### DIFF
--- a/src/msix/common/UnicodeConversion.cpp
+++ b/src/msix/common/UnicodeConversion.cpp
@@ -25,10 +25,10 @@ namespace MSIX {
     {
         if (utf8string.empty()) { return {}; }
         #ifdef WIN32
-        int size = MultiByteToWideChar(CP_ACP, 0, utf8string.data(), static_cast<int>(utf8string.size()), nullptr, 0);
+        int size = MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS, utf8string.data(), static_cast<int>(utf8string.size()), nullptr, 0);
         ThrowLastErrorIf(size == 0, "Error converting to wstring");
         std::wstring result(size, 0);
-        MultiByteToWideChar(CP_ACP, 0, utf8string.data(), static_cast<int>(utf8string.size()), &result[0], size);
+        MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS, utf8string.data(), static_cast<int>(utf8string.size()), &result[0], size);
         #else
         auto converted = utf8_to_utf16(utf8string);
         std::wstring result(converted.begin(), converted.end());
--- a/src/msix/common/UnicodeConversion.cpp
+++ b/src/msix/common/UnicodeConversion.cpp
@@ -25,10 +25,10 @@ namespace MSIX {
     {
         if (utf8string.empty()) { return {}; }
         #ifdef WIN32
-        int size = MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS, utf8string.data(), static_cast<int>(utf8string.size()), nullptr, 0);
+        int size = MultiByteToWideChar(CP_ACP, 0, utf8string.data(), static_cast<int>(utf8string.size()), nullptr, 0);
         ThrowLastErrorIf(size == 0, "Error converting to wstring");
         std::wstring result(size, 0);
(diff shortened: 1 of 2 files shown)


src/msix/common/UnicodeConversion.cpp
```diff
@@ -25,10 +25,10 @@ namespace MSIX {
     {
         if (utf8string.empty()) { return {}; }
         #ifdef WIN32
-        int size = MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS, utf8string.data(), static_cast<int>(utf8string.size()), nullptr, 0);
+        int size = MultiByteToWideChar(CP_ACP, 0, utf8string.data(), static_cast<int>(utf8string.size()), nullptr, 0);
```
DrusTheAxe, line 28: This change needs to be handled differently.

The ActiveCodePage isn't necessarily UTF8 (typically isn't). Changing the implementation of **utf8**_to_wstring(const std::string& **utf8**string) to treat ***utf8***string as the active code page is a breaking change.

You can *add* `string_to_wstring(const std::string& utf8string)` to use CP_ACP, but you open the door to additional questions e.g. should MultiByteToWideChar(...,dwFlags=0) instead of dwFlags=MB_ERR_INVALID_CHARS?

This is probably too low-level a change for the issue. Why is the original string you're handling from a certificate has a non-UTF8 encoding?

jyvenugo: The code that retrieves the publisher name from the certificate is at, https://github.com/microsoft/msix-packaging/blob/e00e737ba572fdedf561b0f246bac56821f95603/src/msix/PAL/Signature/OpenSSL/SignatureValidator.cpp#L285. This is returned as a std::string, which has no implied information on it's encoding at this point.

The original string from the certificate has a non-UTF8 encoding because the certificate is signed with a publisher name which contains non alphanumeric characters in it, for example Île-de-France.

DrusTheAxe: According to [this](https://github.com/ruby/openssl/issues/26#issuecomment-278720820) a certificate's "subject name additional fields" are UTF-8 "[per RFC 4630](https://tools.ietf.org/html/rfc4630#section-3)" 

      The DirectoryString type is defined as a choice of
      PrintableString, TeletexString, BMPString, UTF8String, and
      UniversalString.  CAs conforming to this profile MUST use either
      the PrintableString or UTF8String encoding of DirectoryString,
      with one exception.  When CAs have previously issued certificates
      with issuer fields with attributes encoded using the
      TeletexString, BMPString, or UniversalString, the CA MAY continue
      to use these encodings of the DirectoryString to preserve the
      backward compatibility.

That's dated August 2006. If I'm reading thsi right the only reason subject name wouldn't be UTF-8 is if the cert (a) using PrintableString, TeletexString, BMPString or UniversalString, (b) was issued before Aug'2006 and (c) the CA continued to offer newer replacement certs preserving the old value. Unless this cert's lineage dates back 14+ years then the subject name is not compliant and should be rejected.

If you need to support non-UTF8 values for this legacy reason you'll need a more complex solution than the current one (can't just change UTF8 to ACP - that might fix this one case but creates other (uglier) regressions).

I'm not even sure if it's possible to do something sanely. ACP = Active Code Page. That varies by process (and possibly by thread?). It would be bad for 2 processes to yield different answers for the same cert because they happen to use different ACP (the 'fun' pre-Unicode days of code pages. Shudder).

Where'd the cert come from? Why's it have the value it does? Can it be recreated with the name encoded as UTF-8 (i.e. spec compliant and unambiguous)?

jyvenugo: It appears that the value of the 'publisher' being read from the signature file(certificate) was being read and derived using CertNameToStrA which was returning the ANSI string. I changed it to CertNameToStrW which now returns the UTF8 string.

wcheng-msft (approved):

msftrubengu (approved):

Issue #411, opened 2021-02-01
OPENSSL_add_all_algorithms is not thread safe


sayan-chaliha:
https://github.com/microsoft/msix-packaging/blob/36b3f17f4129ba1e34aee63fb13e8a513f251b56/src/msix/PAL/Signature/OpenSSL/SignatureValidator.cpp#L334

Hello!

I'm using MSIX packages in my application on macOS. Stress testing it for a long duration results in a crash with this trace:
```
  * frame #0: 0x000000019cee5b20 libclang_rt.asan_osx_dynamic.dylib`__asan::AsanDie()
    frame #1: 0x000000019cefc68f libclang_rt.asan_osx_dynamic.dylib`__sanitizer::Die() + 175
    frame #2: 0x000000019cee37c6 libclang_rt.asan_osx_dynamic.dylib`__asan::ScopedInErrorReport::~ScopedInErrorReport() + 422
    frame #3: 0x000000019cee1707 libclang_rt.asan_osx_dynamic.dylib`__asan::ReportDoubleFree(unsigned long, __sanitizer::BufferedStackTrace*) + 231
    frame #4: 0x000000019cedd32b libclang_rt.asan_osx_dynamic.dylib`wrap_free + 267
    frame #5: 0x0000000193efacd5 libmsix.dylib`CRYPTO_free + 37
    frame #6: 0x0000000193f23d01 libmsix.dylib`OBJ_NAME_add + 193
    frame #7: 0x0000000193f1a6e9 libmsix.dylib`OpenSSL_add_all_ciphers + 313
    frame #8: 0x0000000193f1a59e libmsix.dylib`OPENSSL_add_all_algorithms_noconf + 14
    frame #9: 0x0000000193ee34b1 libmsix.dylib`MSIX::SignatureValidator::Validate(IMsixFactory*, MSIX_VALIDATION_OPTION, MSIX::ComPtr<IStream> const&, MSIX::AppxSignatureObject*, MSIX::SignatureOrigin&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >&) + 433
    frame #10: 0x0000000193ecfd30 libmsix.dylib`MSIX::AppxSignatureObject::AppxSignatureObject(IMsixFactory*, MSIX_VALIDATION_OPTION, MSIX::ComPtr<IStream> const&) + 192
[...]

PR #412, opened 2021-02-02
OpenSSL Thread Safety

sayan-chaliha:
* OpenSSL is not thread-safe even when compiled with OPENSSL_THREADS
* Library initialization needs to be called only once and
	* Needs to set a thread ID callback
	* Needs to set locking callback
* OPENSSL_add_all_algorithms() should be called only once in an app
* Addresses [Bug 411](https://github.com/microsoft/msix-packaging/issues/411)

### DIFF
--- a/src/msix/PAL/Signature/OpenSSL/SignatureValidator.cpp
+++ b/src/msix/PAL/Signature/OpenSSL/SignatureValidator.cpp
@@ -12,6 +12,8 @@
 #include <string>
 #include <sstream>
 #include <iostream>
+#include <thread>
+#include <mutex>
 
 #include <openssl/err.h>
 #include <openssl/bio.h>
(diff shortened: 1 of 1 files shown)


src/msix/PAL/Signature/OpenSSL/SignatureValidator.cpp
```diff
@@ -85,6 +87,21 @@ namespace MSIX
         };
     } Asn1Sequence;
 
+    // OpenSSL thread-safety callbacks
+    static void CryptoLockingCallback(std::int32_t mode, std::int32_t n, char const*, std::int32_t) {
```
msftrubengu, line 91: Match repo code style

```
static void CryptoLockingCallback(...)
{
    ...
}
```

sayan-chaliha: Updated!

msftrubengu (approved):

PR #420, opened 2021-02-11
Backport bug fixes into the msix-sdk repo

jyvenugo:
The two bug fixes made by Phil being backported into this repo by this PR are,
-fix memory leak external/msix-sdk/src/msix/PAL/Crypto/Win32/Crypto.cpp
-fix regression in inflatestream.cpp that results in unnecessary memory pressure.

### DIFF
--- a/src/msix/PAL/Crypto/Win32/Crypto.cpp
+++ b/src/msix/PAL/Crypto/Win32/Crypto.cpp
@@ -14,21 +14,60 @@
 
 #include <memory>
 #include <vector>
+#include <deque>
 
-struct unique_hash_handle_deleter {
-    void operator()(BCRYPT_HASH_HANDLE h) const {
-        BCryptDestroyHash(h);
+//////////////////////////////////////////////////////////////////////////////////////////////
+//                                       ScopeGuard                                         //
+//////////////////////////////////////////////////////////////////////////////////////////////
+// inspired by
+// https://channel9.msdn.com/Shows/Going+Deep/C-and-Beyond-2012-Andrei-Alexandrescu-Systematic-Error-Handling-in-C
+class ScopeGuard
+{
+    public:
+    enum class Policy
+    {
+        Always,
+        NoException,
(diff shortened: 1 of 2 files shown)


src/msix/unpack/InflateStream.cpp
```diff
@@ -45,7 +45,7 @@ namespace MSIX {
         {
             ThrowErrorIfNot(Error::InflateRead,(self->m_compressionObject->GetAvailableSourceSize() == 0), "uninflated bytes overwritten");
             ULONG available = 0;
-            self->m_compressedBuffer = std::make_unique<std::vector<std::uint8_t>>(BufferSize);
+            self->m_compressedBuffer.swap(std::make_unique<std::vector<std::uint8_t>>(BufferSize));
```
JohnMcPMS, line 48: It is unclear what value this change is providing, unless there are bugs in the STL/compiler being used.

src/msix/unpack/InflateStream.cpp
```diff
@@ -142,6 +142,8 @@ namespace MSIX {
 
     InflateStream::~InflateStream()
     {
+        m_compressedBuffer = nullptr;
```
JohnMcPMS, line 145: It seems odd that there is memory pressure from having this allocated while `Cleanup` is running.

src/msix/PAL/Crypto/Win32/Crypto.cpp
```diff
@@ -14,21 +14,60 @@
 
 #include <memory>
 #include <vector>
-
-struct unique_hash_handle_deleter {
-    void operator()(BCRYPT_HASH_HANDLE h) const {
-        BCryptDestroyHash(h);
+#include <deque>
```
JohnMcPMS, line 17: I don't agree with the change here; either our usage of unique_ptr is flawed and should be fixed, or this is unnecessary.  Adding this additional overhead isn't the right answer.

PR #424, opened 2021-02-18
Remove TELEMETRYEVENTSOURCE_PUBLIC flag to resolve namespace conflict with external TelemetryEventSource.cs

ryfu-msft:
The TelemetryEventSource from the Msix-packaging library causes a namespace conflict with other projects that want to implement their own managed-code telemetry. By removing this flag and making this class internal, it allows us to create our own telemetry implementation without a namespace conflict.

### DIFF
--- a/tools/utils/Utils/Utils.csproj
+++ b/tools/utils/Utils/Utils.csproj
@@ -23,7 +23,6 @@
     <DebugType>full</DebugType>
     <Optimize>false</Optimize>
     <OutputPath>bin\Debug\</OutputPath>
-    <!--TELEMETRYEVENTSOURCE_PUBLIC is used internally in TelemetryEventSource to make the class public-->
     <DefineConstants>TRACE;DEBUG;TELEMETRYEVENTSOURCE_USE_NUGET</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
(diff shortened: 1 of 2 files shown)


palenshus:
This class gets replaced at release build, so this wouldn't have an effect, there's also other things that are public in this class that could be internal too. Better to remove the flag set here: https://github.com/microsoft/msix-packaging/blob/master/tools/utils/Utils/Utils.csproj#L27

ryfu-msft:
> This class gets replaced at release build, so this wouldn't have an effect, there's also other things that are public in this class that could be internal too. Better to remove the flag set here: https://github.com/microsoft/msix-packaging/blob/master/tools/utils/Utils/Utils.csproj#L27

Reverted changes, and removed defined flag (TELEMETRYEVENTSOURCE_PUBLIC)

tools/utils/Utils/Utils.csproj
```diff
@@ -24,7 +24,7 @@
     <Optimize>false</Optimize>
     <OutputPath>bin\Debug\</OutputPath>
     <!--TELEMETRYEVENTSOURCE_PUBLIC is used internally in TelemetryEventSource to make the class public-->
-    <DefineConstants>TRACE;DEBUG;TELEMETRYEVENTSOURCE_USE_NUGET;TELEMETRYEVENTSOURCE_PUBLIC</DefineConstants>
+    <DefineConstants>TRACE;DEBUG;TELEMETRYEVENTSOURCE_USE_NUGET</DefineConstants>
```
lechacon, line 27: Remove all uses of this. There's another one further down.

lechacon: And update the comment above too, please.

lechacon (approved):

tools/utils/Utils/Utils.csproj
ryfu-msft: removed all TelemetryEventSource_Public flag uses and the unnecessary comment

PR #429, opened 2021-02-26
User/rhuang msft/add file hash to block map

rhuang-msft:
Added \<FileHash> element to block map xml and code to compute the file hash. With this change, every package created by msix.dll will have \<FileHash> element as a child of \<File> element if the uncompressed file size is greater than 64KB. For smaller files, the file hash is the same as block hash as there is only 1 block in the file. So not to add duplicated data and the \<FileHash> element will not be created. Example block map xml with the new \<FileHash> element:
```xml
<BlockMap HashMethod="http://www.w3.org/2001/04/xmlenc#sha256" xmlns="http://schemas.microsoft.com/appx/2010/blockmap" 
     xmlns:b4="http://schemas.microsoft.com/appx/2021/blockmap" IgnorableNamespaces="b4">
          <File Size="189440" Name="App1.exe" LfhSize="38">
          <Block Size="29408" Hash="ORIk+3QF9mSpuOq51oT3Xqn0Gy0vcGbnBRn5lBg5irM="/>
          <Block Size="49487" Hash="BRIk+3QF9mSpuOq51oT3Xqn0jy0vcGbnBRn5lBg5irM="/>
          <Block Size="1487" Hash="CDIk+3QF9mSpuOq51oT3Xqn0jy0vcGbnBRn5lBg5iui="/>
          <b4:FileHash Hash="77hl7hZclsGViUCfuMMqsxsRNJW+PVnNblygNB2vxgE="/>
        </File>
        <File Size="1430" Name="Assets\LockScreenLogo.scale-200.png" LfhSize="65">
          <Block Hash="pBoFOz/DsMEJcgzNQ3oZclrpFj6nWZAiKhK1lrnHynY="/>
      </File>
      ...


</BlockMap>
```

Also changed xml parsing code to disable schema validation when loading block map xml. Block map xml is created by tools and APIs, not by human authors, thus unlikely to be mal-formed.  It can be very big (0.1% of the uncompressed package payload size, so 10MB for 10GB payload) and xml schema validation is CPU intensive and time consuming. We also have semantic validation to catch bad format. We consume what we know and ignore what we don't to make future proof possible. The Windows platform SDK also doesn't validate block map xml schema.

Tested round tripping of  package pack and unpack and verified packages can be packed and unpacked successfully. Verified the create block map xml contains <FileHash> elements as expected. Verified the package containing the new block map xml can be deployed on Windows successfully and app launched successfully.


src/msix/PAL/XML/msxml6/XmlObject.cpp
```diff
@@ -573,17 +573,21 @@ class MSXMLFactory final : public ComClass<MSXMLFactory, IXmlFactory>
...
         #else
         bool HasIgnorableNamespaces = false;
-        NamespaceManager emptyManager;
         #endif
 
         return ComPtr<IXmlDom>::Make<MSXMLDom>(
             stream,
             #if VALIDATING
-                s_xmlNamespaces[static_cast<std::uint8_t>(footPrintType)],
+                // No need to validate block map xml schema as it is CPU intensive, especially for large packages with large block map xml, whihc is
```
msftrubengu, line 587: nit: which

src/inc/internal/Crypto.hpp
```diff
@@ -6,12 +6,45 @@
 
 #include <vector>
 
+#ifndef SHA256_DIGEST_LENGTH
```
msftrubengu, line 9: I am assuming that this can be configurable in the future as a cmake flag in the future?

msftrubengu (approved):

src/inc/internal/Crypto.hpp
rhuang-msft: It is just a precaution to prevent redefinition, given the header is included in different source files in different builds. For example, the same MACRO is defined in openssl's sha.h but not in Win32. With this #ifndef, the same MACRO can be used in both source files.

PR #449, opened 2021-05-13
Adding property on bundle object containing metadata about child packages

palenshus:
Adds new property on `AppxBundleMetadata: ChildAppxPackages`. This replaces the now `[Obsolete]` InternalAppxPackagesRelativePaths, as it has the file paths but also the metadata unique to each child package.

### DIFF
--- a/tools/utils/Utils/AppxPackaging/AppxBundleMetadata.cs
+++ b/tools/utils/Utils/AppxPackaging/AppxBundleMetadata.cs
@@ -137,10 +137,11 @@ namespace Microsoft.Msix.Utils.AppxPackaging
                 else
                 {
                     var childPackageMetadata = new ChildPackageMetadata(
+                        this,
+                        subPackageId.GetPackageFullName(),
                         filePath,
                         subPackageInfo.GetPackageType(),
                         subPackageInfo.GetSize(),
-                        subPackageId.GetPackageFullName(),
                         subPackageId.GetVersion(),
                         subPackageId.GetResourceId());
 
--- a/tools/utils/Utils/AppxPackaging/AppxBundleMetadata.cs
+++ b/tools/utils/Utils/AppxPackaging/AppxBundleMetadata.cs
@@ -57,7 +57,13 @@ namespace Microsoft.Msix.Utils.AppxPackaging
         /// <summary>
         /// Gets the list of all the internal msix packages bundled in this bundle
         /// </summary>
+        [Obsolete]
         public IList<string> InternalAppxPackagesRelativePaths { get; private set; } = new List<string>();
+
+        /// <summary>
+        /// Gets the list of all the internal msix packages bundled in this bundle
+        /// </summary>
+        public IList<ChildPackageMetadata> ChildAppxPackages { get; private set; } = new List<ChildPackageMetadata>();
 
         /// <summary>
         /// Gets the list of all the physical optional msix packages referenced by this bundle, which don't belong to
@@ -114,6 +120,8 @@ namespace Microsoft.Msix.Utils.AppxPackaging
             while (subPackagesEnumerator.GetHasCurrent())
             {
                 IAppxBundleManifestPackageInfo subPackageInfo = subPackagesEnumerator.GetCurrent();
+                IAppxManifestPackageId subPackageId = subPackageInfo.GetPackageId();
+                string filePath = subPackageInfo.GetFileName();
 
                 // If the package is not contained within the bundle, the corresponding package element in the bundle manifest
                 // would not have an offset field. In such a case and only in that case, the AppxPackaging APIs would return 0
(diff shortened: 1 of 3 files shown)


tools/utils/Utils/AppxPackaging/ChildPackageMetadata.cs
```diff
@@ -0,0 +1,75 @@
...
+    {
+        public ChildPackageMetadata(string filePath, APPX_BUNDLE_PAYLOAD_PACKAGE_TYPE packageType, ulong packageSize, string packageFullName, ulong packageVersion, string packageResourceId)
+        {
+            this.FilePath = filePath;
+            this.PackageType = (PackageType)packageType;
+            this.PackageSize = packageSize;
+            this.PackageFullName = packageFullName;
+            this.PackageArchitecture = PackagingUtils.GetPackageArchitectureFromFullName(packageFullName);
+            this.PackageVersion = new VersionInfo(packageVersion);
+            this.PackageResourceId = packageResourceId;
```
lechacon, line 37: Nit: I don't think all properties need to be prefixed with `Package`. I think it would be good to follow the pattern from existing types. Looking at the constructor call and `PackageMetadata` this would be `PackageType`, `Size`, `PackageFullName`, `Architecture`, `Version`, `ResourceId`.

tools/utils/Utils/AppxPackaging/ChildPackageMetadata.cs
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        /// <summary>
+        /// Gets the path the package.
+        /// </summary>
+        public string FilePath { get; }
```
lechacon, line 43: Maybe rename to `RelativeFilePath` for consistency with `ExternalPackageReference.RelativeFilePath` and the obsolete `InternalAppxPackagesRelativePaths`?

tools/utils/Utils/AppxPackaging/AppxBundleMetadata.cs
```diff
@@ -57,8 +57,14 @@ public AppxBundleMetadata(AppxBundleMetadata parentBundleInfo, string fileRelati
         /// <summary>
         /// Gets the list of all the internal msix packages bundled in this bundle
         /// </summary>
+        [Obsolete]
         public IList<string> InternalAppxPackagesRelativePaths { get; private set; } = new List<string>();
 
+        /// <summary>
+        /// Gets the list of all the internal msix packages bundled in this bundle
+        /// </summary>
+        public IList<ChildPackageMetadata> ChildAppxPackages { get; private set; } = new List<ChildPackageMetadata>();
```
lechacon, line 66: Is there a reason to use `Child` instead of `Internal`? Just wondering as this was using Internal/External, but I couldn't find that distinction in the Win32 API. I'm okay with changing it if this is a more appropriate term.

tools/utils/Utils/AppxPackaging/AppxBundleMetadata.cs
```diff
@@ -122,13 +130,25 @@ private void PopulateChildAppxPackages()
...
                         isOptional: false));
                 }
                 else
                 {
-                    this.InternalAppxPackagesRelativePaths.Add(subPackageInfo.GetFileName());
+                    var childPackageMetadata = new ChildPackageMetadata(
+                        filePath,
+                        subPackageInfo.GetPackageType(),
+                        subPackageInfo.GetSize(),
+                        subPackageId.GetPackageFullName(),
```
lechacon, line 143: Nit: Maybe it would be good to follow the same order in both constructors. I.e. parent bundle, package full name, relative file path, and then the others.

lechacon (approved):

tools/utils/Utils/AppxPackaging/AppxBundleMetadata.cs
palenshus: I was influenced by the parent method (PopulateChildAppxPackages), and I also thought Child was better than Internal. To me "Internal" has a connotation of internal code workings, like the internal keyword in c#, and I thought child was more descriptive. Totally fine to change it if you prefer otherwise!

lechacon: That makes sense. If you think "Child" is better than "Internal" let's keep it that way.
Although an external package reference may also be considered a child, hopefully that is not confusing.

tools/utils/Utils/AppxPackaging/ChildPackageMetadata.cs
palenshus: Done

tools/utils/Utils/AppxPackaging/AppxBundleMetadata.cs
palenshus: Done, also added more properties that are on ExternalPackageREference

tools/utils/Utils/AppxPackaging/ChildPackageMetadata.cs
palenshus: Done

lechacon (approved):

PR #453, opened 2021-06-04
Invalid Memory Access in SignatureValidator

sayan-chaliha:
Problem:
* SignatureValidator::IsStoreOrigin() tries to read X.509 extensions
to determine if the origin of the signature matches the Windows Store OID.
* Extension data is converted from a raw buffer to an std::string for
comparision.
* The raw buffer is not null-terminated, and therefore, running
std::strlen() on it causes invalid memory access.
* This invalid access is caught by ASAN on macOS.

Solution:
* Null-terminate the raw buffer before trying to build an std::string
from it.

Tests:
* Ran app test suite that uses libmsix.dylib with ASAN on. No crashes
were reported.

### DIFF
--- a/src/msix/PAL/Signature/OpenSSL/SignatureValidator.cpp
+++ b/src/msix/PAL/Signature/OpenSSL/SignatureValidator.cpp
@@ -131,7 +131,7 @@ namespace MSIX
                             M_ASN1_OCTET_STRING_print(extbio.get(), ext->value);
                         }
                         // null terminate the string.
-                        BIO_write(extbio.get(), "", 1);
+                        BIO_write(extbio.get(), "\0", 1);
                         BUF_MEM *bptr = nullptr;
                         BIO_get_mem_ptr(extbio.get(), &bptr);
                         
--- a/src/msix/PAL/Signature/OpenSSL/SignatureValidator.cpp
+++ b/src/msix/PAL/Signature/OpenSSL/SignatureValidator.cpp
@@ -130,13 +130,12 @@ namespace MSIX
                         {
                             M_ASN1_OCTET_STRING_print(extbio.get(), ext->value);
                         }
-                        // null terminate the string.
-                        BIO_write(extbio.get(), "\0", 1);
+
                         BUF_MEM *bptr = nullptr;
                         BIO_get_mem_ptr(extbio.get(), &bptr);
                         
                         if (bptr && bptr->data && 
-                            std::string((char*)bptr->data).find(OID::WindowsStore()) != std::string::npos)
+                            std::string((char*)bptr->data, bptr->length).find(OID::WindowsStore()) != std::string::npos)
                         {
                             return true;
                         }


sayan-chaliha:
So, somehow, enabling ASAN (via `makemac.sh -asan`) is not catching this issue. I took a look at the CMake files, and it was missing `-fsanitize=address` in linker flags. Modified `src/msix/CMakeLists.txt` and `src/test/msixtest/CMakeLists.txt` to add `-fno-omit-frame-pointer -fsanitize=address` to both compiler and linker commands. The unit tests, though, are still not catching this issue.

src/msix/PAL/Signature/OpenSSL/SignatureValidator.cpp
```diff
@@ -131,7 +131,7 @@ namespace MSIX
                             M_ASN1_OCTET_STRING_print(extbio.get(), ext->value);
                         }
                         // null terminate the string.
-                        BIO_write(extbio.get(), "", 1);
+                        BIO_write(extbio.get(), "\0", 1);
```
JohnMcPMS, line 134: There is no functional difference with your change; an empty string literal is still a char[1] containing a null byte.  You have made a char[2] containing two null bytes.

The function should be returning 1 if it is working correctly (before or after the change).  If it already was, then I don't think that this is the source of the issue.

The better code here is different anyway:
1. Don't write the null byte
2. Construct the string from `bptr->data` and `bptr->length` to avoid the unnecessary strlen call (`std::string_view` would be even better but if I remember correctly this code is C++14)

sayan-chaliha: That makes sense. Making that change now.

msftrubengu:
Can you add your fixes for the CMakeFiles?

sayan-chaliha:
> Can you add your fixes for the CMakeFiles?

So, I verified all the CMake config -- the existing builds with ASAN actually works. I'm not sure why that particular line was not caught by ASAN in MSIX builds.

I tested by:
* Removing the write of the null-byte: ASAN catches it.
* Using a permutation of the linker and compiler flags: ASAN works with the existing config.

Somehow, Office's builds flag it as an issue, while running `msixtest` with ASAN doesn't. Using `bptr->length` works for both Office and `msixtest` with ASAN.

sayan-chaliha:
> > Can you add your fixes for the CMakeFiles?
> 
> So, I verified all the CMake config -- the existing builds with ASAN actually works. I'm not sure why that particular line was not caught by ASAN in MSIX builds.
> 
> I tested by:
> 
> * Removing the write of the null-byte: ASAN catches it.
> * Using a permutation of the linker and compiler flags: ASAN works with the existing config.
> 
> Somehow, Office's builds flag it as an issue, while running `msixtest` with ASAN doesn't. Using `bptr->length` works for both Office and `msixtest` with ASAN.

May have some thing to do with the ASAN lib used in Office builds specifically.

JohnMcPMS (approved):

PR #477, opened 2021-09-14
Use MS-DOS time in LFH and CDH

### LINKED ISSUE #393: Files extracted after package/bundle creation have the date modified column show up as 5/17/2061
Steps:
Create a package/bundle using makemsix pack or bundle command.
Right click on the created package and rename to zip.
Extract all the contents.
The 'Date modified' field appears as 5/17/2061 when it should be the current date.

msftrubengu:
Fix #393 

For convenience, we use some magic number in the last modify date and time in the LFH and CDH. This produce the behavior described by #393.

This change gets the local system time and convert it to MS-DOS date and time format and add it to the headers. It also follows the makeappx implementation that if getting the time fails the default is 12/31/2107, 23:59:58.

Validated in Windows and macOS.

### DIFF
--- a/src/inc/shared/TimeHelpers.hpp
+++ b/src/inc/shared/TimeHelpers.hpp
@@ -1,0 +1,28 @@
+//
+//  Copyright (C) 2021 Microsoft.  All rights reserved.
+//  See LICENSE file in the project root for full license information.
+// 
+#pragma once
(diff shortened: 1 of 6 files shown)


ranm-msft (approved):
:shipit:

PR #480, opened 2021-09-21
Remove unused openssl submodules

### LINKED ISSUE #476: openssl has been imported into the tree, with submodules, incorrectly
So, msix-packaging has a "subtree" of openssl, which seems to be a direct import of the entire content of the openssl repository verbatim, but with paths mapped. The problem is that openssl uses git submodules, and these submodules--which are represented with special entries in the git tree object--were also imported verbatim into the repository... with remapped paths, but without correctly importing the .gitmodules into the top-level folder.

```
# git cat-file -p HEAD:lib/openssl | grep commit
160000 commit 2070f8ad9151dc8f3a73bffaa146b5e6937a583f  boringssl
160000 commit 890ca2f401924cdcb88f493950b04700bbe52db3  krb5
160000 commit 09403100de2f6f1cdd0d484dcb8e620f1c335c8f  pyca-cryptography
```

The result is that if you attempt to do a `git submodule init` inside of the msix-packaging working copy, you get an error.

```
# git submodule init
fatal: No url found for submodule path 'lib/openssl/boringssl' in .gitmodules
```

You might be saying "but our project doesn't use submodules"... but a lot of git tooling expects that if submodules exist that they are registered correctly, and if _someone else_ (such as me) wants to put your repository as a submodule of _their_ repository, they are going to get errors attempting to do the standard `git submodule update --init --recursive` (which I would argue is something that should be so safe you always do it after any checkout).

```
# git submodule update --init --recursive
fatal: No url found for submodule path 'app-windows/msix-packaging/lib/openssl/boringssl' in .gitmodules
Failed to recurse into submodule path 'app-windows/msix-packaging'
```

As I'm pretty sure you aren't actually using these submodules, I'd suggest that the correct fix is to simply drop these commit entries from your tree objects when you do the merging in of the openssl repository into your tree.

msftrubengu:
Removing the three submodules of the openssl subtree as we don't use them.

Resolves #476


JohnMcPMS (approved):

ranm-msft (approved):
:shipit:

PR #502, opened 2022-02-03
Add code security analysis tasks to AzDO Extension build pipelines

lechacon:
Changes to the build pipelines for the AzDO extension:
* Added CredScan and Semmle code security analysis tasks.
* Factored out the common steps between the PR and release build pipelines.
* Moved to use a custom agent pool.


JohnMcPMS:
I think the fact that it is still queued after 5 hours suggests that something is not quite right...

lechacon:
> I think the fact that it is still queued after 5 hours suggests that something is not quite right...

Oh, right. I needed to give permissions to the pipeline to use the agent pool. Just did that, so it shouldn't be a problem in the future.

msftrubengu (approved):

PR #505, opened 2022-03-01
Add definition of pipeline to create NuGet packages for MSIX SDK

lechacon:
Adding a definition of a pipeline to produce NuGet packages for the MSIX SDK.

Moved the definitions of each platform-specific job in the build pipelines to a template which is used by the respective pipeline, and added a new pipeline that uses all of these jobs and then creates the packages in a new stage. The new stage is based on an existing internal release pipeline, with minor changes (replaced build artifacts by pipeline artifacts, enabled generation for Linux, added a step to produce the SBoM manifest)


pipelines/azure-pipelines-nuget.yml
```diff
@@ -0,0 +1,388 @@
...
+      displayName: 'Archive files to sign'
+      inputs:
+        rootFolderOrFile: ToSign
+        includeRootFolder: false
+        archiveFile: 'Signed\artifacts.zip'
+
+    - task: SFP.build-tasks.custom-build-task-1.EsrpCodeSigning@1
+      displayName: 'ESRP CodeSigning macOS dylibs'
+      inputs:
+        ConnectedServiceName: 'ESRP Connection MsixSDK 2'
```
msftrubengu, line 172: >ESRP Connection MsixSDK 2 [](http://example.com/codeflow?start=31&length=25)

We should make this a build parameter, so when we have to renew it we don't have to create new PR.

pipelines/azure-pipelines-nuget.yml
```diff
@@ -0,0 +1,388 @@
...
+      inputs:
+        command: pack
+        packagesToPack: 'package_mac\Package.nuspec'
+        packDestination: '$(NuGetsFolder)\macos'
+
+    # Windows
+    - task: SFP.build-tasks.custom-build-task-1.EsrpCodeSigning@1
+      displayName: 'ESRP CodeSigning Windows'
+      inputs:
+        ConnectedServiceName: 'ESRP CodeSigning connection 4 windows'
```
msftrubengu, line 276: >ESRP CodeSigning connection 4 windows [](http://example.com/codeflow?start=31&length=37)

same with windows

pipelines/azure-pipelines-nuget.yml
```diff
@@ -0,0 +1,388 @@
...
+      displayName: 'Prepare Windows'
+
+    - task: NuGetCommand@2
+      displayName: 'NuGet pack Windows'
+      inputs:
+        command: pack
+        packagesToPack: 'package_win\Package.nuspec'
+        packDestination: '$(NuGetsFolder)\windows'
+
+    # Global tasks
```
msftrubengu, line 352: ># Global tasks [](http://example.com/codeflow?start=4&length=14)

Can you add a comment that this is just a fail safe to make sure the artifacts come from the same build

pipelines/azure-pipelines-nuget.yml
lechacon: Why would we have to create a new PR? I'd think we could just edit the existing connection.

PR #519, opened 2022-06-22
Fix memory leaks ZipObject.cpp

sayan-chaliha:
* CentralDirectoryFileHeader and LocalFileHeader uses MsDosDateAndTime.
* MsDosDateAndTime is allocated (`new`) and not deallocated (`delete`).
* MsDosDateAndTime doesn't need to be allocated in the heap.


msftrubengu:
I looked into it, and AOSP tests started failing early June.  I'm going to disable the test to unblock you.

sayan-chaliha:
Thanks Ruben!

PR #525, opened 2022-07-11
Update Microsoft.NET.Test.Sdk NuGet dependency on MSIX.Utils tests

lechacon:
Per https://github.com/advisories/GHSA-5crp-9r3c-p9vr, we should update the dependency we had on Newtonsoft to at least v13.0.1. We had it as a transitive dependency from Microsoft.NET.Test.Sdk; so we had to update that. The new version of Newtonsoft is used starting on the latest v17.4 of Microsoft.NET.Test.Sdk.

This was only present in the test code, so we won't need to make a new release of the Utils.

### DIFF
--- a/tools/utils/UtilsTests/UtilsTests.csproj
+++ b/tools/utils/UtilsTests/UtilsTests.csproj
@@ -38,9 +38,9 @@
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


ranm-msft (approved):
:shipit:

PR #565, opened 2023-02-23
Get code building under c++20

Penguinwizzard:
The only thing that seems to break when building under c++20 is that the u8-string literals change to char8_ts. However, we're not using any character in these literals that would require the support, and the rest of the code isn't using the right functions to have robust unicode support anyway, so I swapped to just classic char-string-literals.

I did a drive-by improvement to allow for some automatic determination of the right functions to use for these lists, if someone does want to change the types back to u8-literals (so, char under c++14 and char8_t under c++20), but there's issues further throughout the code that uses these values that building under c++20 reveals; I suspect that we're not universally using the right string comparison functions.

### DIFF
--- a/makelinux.sh
+++ b/makelinux.sh
@@ -16,6 +16,7 @@ usage()
     echo $'\t' "--pack                  Include packaging features. Sets validation parser on."
     echo $'\t' "--skip-samples          Skip building samples."
     echo $'\t' "--skip-tests            Skip building tests."
+    echo $'\t' "--extra                 Extra flags to CMake."
 }
 
 printsetup()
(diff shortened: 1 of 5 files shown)


makelinux.sh
```diff
@@ -16,6 +16,7 @@ usage()
     echo $'\t' "--pack                  Include packaging features. Sets validation parser on."
     echo $'\t' "--skip-samples          Skip building samples."
     echo $'\t' "--skip-tests            Skip building tests."
+    echo $'\t' "--extra                 Extra flags to CMake."
```
msftrubengu, line 19: My concern that extra ends up as a place to dump all the other new good thing in the future.

Can you make as something like -cxxstd 20 with default value of 14 and then just pass -DCMAKE_CXX_STANDARD=<value> to cmake?

Penguinwizzard: Sure, I've switched to a pattern like that.

PR #579, opened 2023-04-28
User/masudars/update openssl 1 1 1t

Madhusudhan-MSFT:
CG Fix for OpenSSL. The fix includes

1. Removing old openssl 1_1_1j  subtree directory from the repo folder
2. Re-add openssl 1_1_1t  subtree that contains security fix

`''git subtree add --prefix lib/openssl https://github.com/openssl/openssl.git OpenSSL_1_1_1t --squash''`

**[How Validated:]**
	1. makewin.cmd x64
		a. .vs\msixtest\msixtest.exe
                     All tests passed (11292 assertions in 72 test cases)
	2. makewin.cmd x64 -px
		a. .vs\msixtest\msixtest.exe
                     All tests passed (11292 assertions in 72 test cases)

But windows platform doesn't use openssl and I am relaying build results from mac/ios build pipeline as linux build pipeline is broken.


msftrubengu:
makewin.cmd x64 doesn't use OpenSSL but the windows crypo APIs. Try `makewin.cmd x64 --crypto-openssl` or set up WSL for use makelinux.

Also, since we don't have CI for Windows using OpenSSL I suggest you do that in master first to confirm it builds, then try it with the new subtree.

Madhusudhan-MSFT:
Looks like there are some breaking changes probably new file additions what is causing validation pipeline failure. I will some time to analyze what is causing compilation failure and update the fix.

I wil also try ( if that makes it any easier for the fix the failures)

makewin.cmd x64 --crypto-openssl or set up WSL for use makelinux.

Madhusudhan-MSFT:
**msix-packaging Linux CI** - failure seems to be due to infra issue as it fails with below error.

```
##[warning]An image label with the label ubuntu-18.04 does not exist.
,##[error]The remote provider was unable to process the request.
```

msftrubengu:
Can you open an issue to update the linux image? 

Also since you are updating the subtree, don't squash this PR

msftrubengu (approved):

Madhusudhan-MSFT:
I opend #580 to track linux image update.

PR #600, opened 2023-10-06
User/smohamed/update to support logging improv

Smohamed-MSFT:


Smohamed-MSFT:
Validated by reverting to previous commit as the current wasn't building. Successfully output the errors to a log file when specified using the -output flag.

MsixCore/msixmgr/CommandLineInterface.cpp
```diff
@@ -175,6 +175,19 @@ std::map<std::wstring, Options, CaseInsensitiveLess> CommandLineInterface::s_opt
                     return S_OK;
                 }),
             },
+            {
+                L"-output",
+                Option(true, IDS_STRING_HELP_OPTION_OUTPUT,
```
sreading-MSFT, line 180: >IDS_STRING_HELP_OPTION_OUTPUT [](http://example.com/codeflow?start=29&length=29)

Since this option is unpack only I'd rename these to fit the unpack specific options elsewhere, so IDS_STRING_HELP_OPTION_UNPACK_OUTPUT

MsixCore/msixmgr/CommandLineInterface.cpp
```diff
@@ -175,6 +175,19 @@ std::map<std::wstring, Options, CaseInsensitiveLess> CommandLineInterface::s_opt
...
+            {
+                L"-output",
+                Option(true, IDS_STRING_HELP_OPTION_OUTPUT,
+                    [&](CommandLineInterface* commandLineInterface, const std::string& outPath)
+                {
+                    if (commandLineInterface->m_operationType != OperationType::Unpack)
+                    {
+                        return E_INVALIDARG;
+                    }
+                    commandLineInterface->m_outputPath = utf8_to_utf16(outPath);
```
sreading-MSFT, line 187: >m_outputPath [](http://example.com/codeflow?start=42&length=12)

Would rename this to m_unpackOutputPath as well.

sreading-MSFT: Actually is "output" set in stone? Without reading the help I'm not sure it would be obvious what the difference between "-output" and "-destination" would be. Maybe m_outputLoggingPath would be more clear at least in code if the cmdline interface arg name is set.

MsixCore/msixmgr/msixmgr.cpp
```diff
@@ -801,7 +871,15 @@ int main(int argc, char * argv[])
...
 
-                    OutputUnpackFailures(packageSourcePath, skippedFiles, failedPackages, failedPackagesErrors);
+                    if (outputFilePath.empty())
+                    {
+                        OutputUnpackFailures(packageSourcePath, skippedFiles, failedPackages, failedPackagesErrors);
+                    }
+                    else
+                    {
+                        OutputUnpackFailuresToFile(packageSourcePath, skippedFiles, failedPackages, failedPackagesErrors, outputFilePath);
+                        wcout << "Written error output to file: " << outputFilePath;
```
sreading-MSFT, line 881: >Written [](http://example.com/codeflow?start=34&length=7)

Wrote

MsixCore/msixmgr/msixmgr.rc
```diff
@@ -134,7 +134,7 @@ BEGIN
 	IDS_STRING_HELP_OPTION_MOUNT_FILETYPE "the type of file to mount or unmount. The following file types are currently supported: {VHD, VHDX, CIM}"
 	IDS_STRING_HELP_OPTION_UNPACK_VHDSIZE "the desired size of the VHD or VHDX file in MB. Must be between 5 and 2040000 MB. Use only for VHD or VHDX files"
 	IDS_STRING_HELP_OPTION_MOUNT_READONLY "boolean (true of false) indicating whether a VHD(X) should be mounted as read only. If not specified, the image is mounted as read-only by default"
-    
+	IDS_STRING_HELP_OPTION_OUTPUT "optional parameter selects logging location to write the command line output to"
```
sreading-MSFT, line 137: > to write the command line output to [](http://example.com/codeflow?start=75&length=36)

Code as implemented does not write all command line output, only the errors. I assume this is intentional since piping all output to a file is a built in feature of powershell so probably does not need to be a special parameter here. Would specify that this is errors only (and maybe only a very specific set of errors at that? Is it supposed to be writing all errors or just the list of files that failed?)

MsixCore/msixmgr/msixmgr.cpp
```diff
@@ -801,7 +870,15 @@ int main(int argc, char * argv[])
                     std::wcout << "Finished unpacking packages to: " << unpackDestination << std::endl;
                     std::wcout << std::endl;
 
-                    OutputUnpackFailures(packageSourcePath, skippedFiles, failedPackages, failedPackagesErrors);
+                    if (outputFilePath.empty())
```
sreading-MSFT, line 874: >  if (outputFilePath.empty()) [](http://example.com/codeflow?start=18&length=29)

Not sure it makes sense to have this be an either\or case. As long as so much other logging is still going to the command line, it seems like this stuff should still go there too even if output path is set. If outputpath is set it can be duplicated to the file.

MsixCore/msixmgr/msixmgr.cpp
```diff
@@ -173,6 +173,60 @@ void OutputUnpackFailures(
...
 }
 
+void OutputUnpackFailuresToFile(
+    _In_ std::wstring packageSource,
+    _In_ std::vector<std::wstring> skippedFiles,
+    _In_ std::vector<std::wstring> failedPackages,
+    _In_ std::vector<HRESULT> failedPackagesErrors,
+    _In_ std::wstring outfilePath)
+{
+    std::wofstream  outfile;
```
sreading-MSFT, line 184: >std::wofstream [](http://example.com/codeflow?start=4&length=14)

Duplicating all this code for writing to the different types of logs should be unnecessary, but writing multiple streams at once also seems to have some complexity. Easiest solution seems like building a big wstringstream of everything that needs to get logged in this function and then having the wrapper function either write that to just std::wcout or both std::wcout and the file stream.

std::wstringstream unpackFailureString;
use unpackFailureString everywhere std::wcout is used.
add unpackFailureString << std::flush;

then do:
    std::wcout << unpackFailureString.rdbuf() << std::endl;
if (!outputFilePath.empty)
{
open
   outFile << unpackFailureString.rdbuf() << std::endl;
}

sreading-MSFT:
>         std::wcout << std::endl;

At a very high level I'm trying to figure out if the feature this code review is trying to implement would work better by just changing all the std::wcout uses in this particular function to std::wcerr and then just let the caller pipe the error output to a file rather than having a separate command line arg.

---
Refers to: MsixCore/msixmgr/msixmgr.cpp:135 in 2ce316d. [](commit_id = 2ce316dc7a499052bb6bac3ea6ef314a40d1e304, deletion_comment = False)

MsixCore/msixmgr/msixmgr.cpp
Smohamed-MSFT: If output path is set this would end up being logged in the file

MsixCore/msixmgr/msixmgr.cpp
Smohamed-MSFT: The duplication works because we want to have different behavior when output path is set and when it isn't this is something I've validated with PM during design.

MsixCore/msixmgr/CommandLineInterface.cpp
Smohamed-MSFT: fixed

MsixCore/msixmgr/CommandLineInterface.cpp
Smohamed-MSFT: Sure nice catch. Fixed!

MsixCore/msixmgr/msixmgr.cpp
Smohamed-MSFT: Fixed

MsixCore/msixmgr/msixmgr.rc
Smohamed-MSFT: That is correct

MsixCore/msixmgr/msixmgr.cpp
sreading-MSFT: Duplicating all the output string logic isn't necessary to have different behavior when output path is set.


void GetUnpackOutputFailures(
    _In_ std::wstring packageSource,
    _In_ std::vector<std::wstring> skippedFiles,
    _In_ std::vector<std::wstring> failedPackages,
    _In_ std::vector<HRESULT> failedPackagesErrors,
    _In_ std::wstringstream& unpackFailureStringStream)
{
    if (!skippedFiles.empty())
    {
        unpackFailureStringStream << std::endl;
        unpackFailureStringStream << "[WARNING] The following items from " << packageSource << " were ignored because they are not packages or bundles " << std::endl;
...
etc
}

void OutputUnpackFailures(
    _In_ std::wstring packageSource,
    _In_ std::vector<std::wstring> skippedFiles,
    _In_ std::vector<std::wstring> failedPackages,
    _In_ std::vector<HRESULT> failedPackagesErrors,
    _In_ std::wstring outputFilePath)
{
    std::wstringstream unpackFailureStringStream;
    GetUnpackOutputFailures(packageSource, skippedFiles, failedPackages, failedPackagesErrors, unpackFailureStringStream);
    std::wstring unpackFailureString = unpackFailureStringStream.str();
    std::wcout << unpackFailureString << std::endl;
    if (!outputFilePath.empty())
    {
        std::wofstream  outFile;
        outFile.open(outputFilePath);
        outFile << unpackFailureString << std::endl;
        outFile.close();
        wcout << "Wrote error output to file: " << outputFilePath;
    }
}

And then just call the OutputUnpackFailures function and let it deal with the if\else for if outputFilePath is set.
std::wcout << std::endl;
std::wcout << "Finished unpacking packages to: " << unpackDestination << std::endl;
std::wcout << std::endl;

OutputUnpackFailures(packageSourcePath, skippedFiles, failedPackages, failedPackagesErrors, outputFilePath);

Smohamed-MSFT: Completed making those changes https://github.com/microsoft/msix-packaging/pull/600/commits/82a4c90d06d399891ba0dcfa67ab71904f1b30e9

sreading-MSFT:
Looks like my username isn’t marked as valid for approving on this project for checkin - but thank you for all the updates. I approve.

PR #604, opened 2023-10-28
Fix compilation on gcc13

deepbluev7:
Fairly straightforward. gcc13 cleaned up some headers, so you need to manually include them nowadays.


deepbluev7:
@microsoft-github-policy-service agree

PR #612, opened 2024-01-11
Update ESRP Codesigning task version in ado pipelines

npuvvada:
## Why is this change being made?
ESRP code signing failure with current ESRP codesigning task version 1.0.
The .NET dependencies installed on the agent machine do not match with the .NET dependencies required for ESRP codesigning earlier task version. 

## What changed?
ESRP Code signing task version is changed to 4.0

## How was the change tested?
Verified ESRP code signing task.


rahul89dadhich (approved):
looks good

PR #651, opened 2024-10-22
Updated version of Signing changes

Penguinwizzard:
This is a rebase of johnmcpms/signing on the latest version of microsoft/msix-packaging@master, with a couple of small patches to get it working (and handling pfx passwords). Ideally, with a simpler history and a few of the merge fixes rolled in, this should be an easier change to land than the existing branches where multiple contributors have done repeated merges of master into the topic branch.


src/inc/internal/ObjectBase.hpp
```diff
@@ -19,7 +21,10 @@ namespace MSIX { namespace Meta {
 // there is exactly one value that this field is allowed to be
 template <typename T>
 static void ExactValueValidation(T value, T spec) {
-    ThrowErrorIfNot(Error::InvalidParameter, spec == value, "Incorrect value specified at field.");
+    if (spec != value) {
+        std::cout << "Value in field doesn't match expection; expected " << spec << ", got " << value << std::endl;
```
JohnMcPMS, line 25: Rather than direct output to `std::cout`, I would think it better to construct the string and include it in the thrown exception.

src/makemsix/main.cpp
```diff
@@ -635,6 +635,7 @@ Command CreateSignCommand()
             Option{ "-p", "Package file path.", true, 1, "package" },
             Option{ "-c", "Certificate file path.", true, 1, "cert" },
             Option{ "-cf", "Certificate format.", false, 1, "format" },
+            Option{ "-pass", "Certificat epassword.", false, 1, "password" },
```
JohnMcPMS, line 638:
```suggestion
            Option{ "-pass", "Certificate password.", false, 1, "password" },
```
Typo

src/msix/PAL/Crypto/OpenSSL/SignatureCreator.cpp
```diff
@@ -35,7 +36,7 @@ namespace MSIX
                     unique_BIO certBIO{ BIO_new_mem_buf(reinterpret_cast<void*>(certBytes.data()), static_cast<int>(certBytes.size())) };
                     unique_PKCS12 cert{ d2i_PKCS12_bio(certBIO.get(), nullptr) };
 
-                    ParsePKCS12(cert.get(), nullptr, "Unable to open PFX file");
+                    ParsePKCS12(cert.get(), pass, "Unable to open PFX file - incorrect password?");
```
JohnMcPMS, line 39: `ThrowOpenSSLErrIfFailedMsg` should be extracting useful information from OpenSSL about the nature of the error.

src/makemsix/main.cpp
```diff
@@ -594,6 +628,52 @@ Command CreatePackCommand()
...
+        {
+            std::cout << "WARNING: The signing feature is not complete, see the help for this command for more information." << std::endl;
+            std::cout << std::endl;
+
+            return SignPackage(
+                MSIX_SIGNING_OPTIONS::MSIX_SIGNING_OPTIONS_NONE,
+                const_cast<char*>(invocation.GetOptionValue("-p").c_str()),
+                GetCertificateFormat(invocation),
+                const_cast<char*>(invocation.GetOptionValue("-c").c_str()),
+                const_cast<char*>(invocation.GetOptionValue("-pass").c_str()),
```
JohnMcPMS, line 670: The OpenSSL docs specify the encoding of the password, but our `main` does not attempt to enforce any specific type of encoding on the input.  This would need to be addressed for full support.

JohnMcPMS (commented):
Reviewed only the changes by @Penguinwizzard , as it seems unhelpful to review my own changes even if I am 5 years removed from them.  My memory is that these changes were working on the golden path but lacked any automated tests.  That time requirement was the reason they never got checked in.  Ultimately I think that this still needs to be addressed to merge.

src/msix/PAL/Crypto/OpenSSL/SignatureCreator.cpp
Penguinwizzard: I've changed this to not add the extra text, and instead leave it to the OpenSSL message.

Penguinwizzard:
I've now added some automated tests, albeit fairly simple ones - they pack an MSIX, sign it, and then validate its signature and contents.

src/makemsix/main.cpp
Penguinwizzard: While that _is_ true, I think that the issue exists more broadly in consistently supporting various encodings in the command-line flags in makemsix. Addressing that fits better in the scope of a follow-up PR to handle it for both the case of passwords, as here, and also for the variety of cases that come up with filenames across OSes.

PR #657, opened 2025-01-08
Bring modifications made during MSFT repoman project

ryknuth:
### Change 1:
Defer evaluation IAppxFile's underlying stream

During the client MSIX scenario, when deploying via the DO transport, we were seeing that effectively the whole client MSIX file was being streamed to the local device. After a bunch of debugging, we figured out that the culprit to this is that the MSIX SDK is validating that every file (whether read or not) that's specified within the central directory is also covered by the blockmap, AND that the size recorded in the blockmap is the size recorded in the underlying OPC structures. This sounds great and all, but there's two problems, 1) in order to properly validate the OPC's metadata w.r.t. a file, a structure at the beginning of each file must be read from the OPC stream before the zip interface can answer how big a particular part of the file is. and 2) we end up doing this validation for EVERY file, whether its actually required for the deployment or not; which is HUGELY inefficient w.r.t. not only memory, but I/O.

The fix then, is to lazily perform this sort of validation, delaying it until the caller actually requests the underlying stream for a corresponding IAppxFile. This ensures that before a file's contents are accessed, it's metadata aligns with what's specified in the blockmap, but also, it ensures that the I/O on the underlying file's metadata in the OPC structure (which sequentially comes just before the file) is done right before the file's contents are most likely to be read.

### Change 2:
Http transport usage is terrible

The change reduces the rather verbose/chatty usage of the underlying stream in the zip reader layer of the MSIX SDK to being a little bit smarter. Basically, when we know how big a structure is, rather than reading it piecemeal (e.g. 2, 4, or 8 bytes at a time), we calculate the whole structure's size, and read those bytes all at once into a stack-local buffer, and then use that in-memory copy to perform the remaining serialization.

The biggest part of the change is when it comes to reading the zip central directory. Since each and every one of those structs are variable size, but they all MUST come sequentially, one after the other, from the start of the 1st instance all the way to the zip footer, we can read that entire sequence into memory and use the existing "small read/chatty/verbose" pattern, as that's going to consume what's in-memory as opposed to dozens of small reads times potentially hundreds to thousands of files.

### DIFF
--- a/src/inc/internal/AppxFile.hpp
+++ b/src/inc/internal/AppxFile.hpp
@@ -23,14 +23,9 @@ namespace MSIX {
     class AppxFile : public ComClass<AppxFile, IAppxFile, IAppxFileUtf8>
     {
     public:
-        AppxFile(IMsixFactory* factory, const std::string& name, const ComPtr<IStream>& stream) : m_factory(factory), m_name(name), m_stream(stream)
-        {
-            LARGE_INTEGER start = { 0 };
-            ULARGE_INTEGER end = { 0 };
-            ThrowHrIfFailed(m_stream->Seek(start, StreamBase::Reference::END, &end));
-            ThrowHrIfFailed(m_stream->Seek(start, StreamBase::Reference::START, nullptr));
-            m_size = end.u.LowPart;
-        }
+        AppxFile(IMsixFactory* factory, const std::string& name, std::function<ComPtr<IStream>()>&& streamFunc)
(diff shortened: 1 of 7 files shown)


ryknuth:
@msftphil MSIX changes. Please review.

msftphil:
Yes, this looks like the correct change to bring to support efficient streaming AppX/MSIX packages over an HTTP transport.

msftphil (approved):
LGTM

msftrubengu (approved):

PR #675, opened 2025-05-20
Fix case sensitive includes for `unknwn.h` and `objidl.h`

JanisErdmanis:
Windows and MacOS are both case-insensitive, and hence the issue of wrong capitalisation may not have surfaced.

I am forming a recipe for cross-compilation for the Julia BinarBuilder environment, which uses Linux x86_64 as the host system. However, as it uses a sensitive file system, I encountered a linking error that could only be resolved by lowering cases for `unknwn.h` and `objidl.h`.

More on that can be found in the pull request:
https://github.com/JuliaPackaging/Yggdrasil/pull/10944


JanisErdmanis:
@microsoft-github-policy-service agree

PR #708, opened 2026-05-15
Fix `make install` failing after CMAKE_INSTALL_PREFIX is changed

pvutov:
The zlib subproject is setting several unnecessary cache variables: `INSTALL_BIN_DIR` etc. The values for these variables are only derived once, during the initial configuration, from CMAKE_INSTALL_PREFIX.

The specific bug this causes is:

```
cmake -DCMAKE_INSTALL_PREFIX=/tmp/install1 ..
make install # works
cmake --install . --prefix /tmp/install2  # does not do what we'd expect, because INSTALL_BIN_DIR is not properly updated
```

The first commit on my fork merely fixes the bug. The second commit also fixes an IMO bad design decision by removing the unnecessary cache variables involved in the bug.


pvutov:
@microsoft-github-policy-service agree company="RUAG AG"

pvutov:
I am noticing that this project more generally does not support installing; the main msix-packaging binary and lib do not show up in the install folder even with my fix. But that can be fixed separately by adding an `install(` command to `src/makemsix/CMakeLists.txt` :sweat_smile:

